PR #11, opened 2020-12-09
Fix error formatting (missing a call to Sprintf)

stjepang:
Just a minor bug I noticed while skimming over the code...


ripienaar:
nice, thanks

PR #36, opened 2021-01-22
Add replicas handling in Stream editing

rytswd:
It seems `nats str edit` with new replica count was not handled. It may be better to add some validation, but kept it simple for now, as I assume the server will validate the input.

### DIFF
--- a/nats/stream_command.go
+++ b/nats/stream_command.go
@@ -669,6 +669,10 @@ func (c *streamCmd) copyAndEditStream(cfg api.StreamConfig) (api.StreamConfig, e
 		cfg.Duplicates = dw
 	}
 
+	if c.replicas != 0 {
+		cfg.Replicas = int(c.replicas)
+	}
+
 	return cfg, nil
 }
 


nats/stream_command.go
```diff
@@ -669,6 +669,10 @@ func (c *streamCmd) copyAndEditStream(cfg api.StreamConfig) (api.StreamConfig, e
 		cfg.Duplicates = dw
 	}
 
+	if c.replicas != 0 {
+		cfg.Replicas = int(c.replicas)
```
ripienaar, line 673: I don't believe we can edit replica count, we should error if someone attempts to edit this for the moment.

rytswd: Oh OK, if that's the case, it may be better to remove from the help, if feasible?

ripienaar: Hmm, I have a function shared between a few sub commands to add all the flags, can maybe remove yes - personally I want to allow replicas to be edited I think its important but right now its not implemented on the server, the admin/user/management of clustered JS is a bit up in the air still both in implementation and what we'll support.

rytswd: Couldn't find a simple solution but to add replicas flag handling to `add`, `copy` and `create` specifically
https://github.com/nats-io/natscli/compare/main...rytswd:drop-edit-support-for-replicas?expand=1

Should I close this PR and create a separate one with the above change?

ripienaar: Lets keep it open, I dont really know what the right answer is here - we might yet support it.  So lets leave it open as is as a reminder for when we have figured out what we want?

rytswd: Sorry I didn't notice I was cross-commenting to your previous one, too - and I agree, it would be great if I would definitely like to be able to edit the replication count. I was specifically trying to test where I start with a few nodes and add more nodes to the cluster. Recreating the Stream is one option, but probably not ideal for most of use cases IMO.

ripienaar: Yeah exactly, I think we need to assist with maintenance windows too, so we'll see, we'll revisit the CLI limits and controls once we have all that bedded down.

As part of https://github.com/nats-io/jetstream/issues/414 we'll look at making sense of how it all goes, lets keep this one open but not change now

ripienaar:
We now support this in the Server which was not supported when you initially opened the PR. We also updated the CLI to support this in the manner the server supports, I will close this PR, thanks for the contribution

PR #93, opened 2021-02-28
use a duration for expires

ripienaar:


ripienaar:
Reference https://github.com/nats-io/nats-server/pull/1944/

PR #146, opened 2021-04-23
Add --max-pending suggestion to help

bmidgley:
A quick fix to make the docs help people find the flag name


ripienaar:
Thanks a lot

PR #176, opened 2021-06-16
Add a --tls-dump-certs flag

philpennock:


nats/main.go
```diff
@@ -84,6 +85,7 @@ See 'nats cheat' for a quick cheatsheet of commands
 	ncli.Flag("domain", "JetStream domain to access").PlaceHolder("PREFIX").PlaceHolder("DOMAIN").Hidden().StringVar(&jsDomain)
 	ncli.Flag("context", "Configuration context").Envar("NATS_CONTEXT").StringVar(&cfgCtx)
 	ncli.Flag("trace", "Trace API interactions").BoolVar(&trace)
+	ncli.Flag("tls-dump-certs", "Dump TLS first certificate chain for server").BoolVar(&tlsDumpCerts)
```
ripienaar, line 88: not sure this should be on the global set of flags - we already have quite a lot and I'd rather not add ones that not strictly applicable.

I can think of a few optoins here:

 * We add it to the `nats account info` which already shows a log of connection things like IPs and ports
 * We add it to context show/validate

Probably I prefer the first one

```
[rip@p1-lon]% nats account info
Connection Information:

               Client ID: 2560
               Client IP: 46.101.44.80
                     RTT: 688.062µs
       Headers Supported: true
         Maximum Payload: 1.0 MiB
       Connected Cluster: lon
           Connected URL: nats://n3-lon:4222
       Connected Address: 188.166.173.133:4222
     Connected Server ID: NBFWZTPM3AVUN4S6LONLHO4SQOMCCREHDJ5SPDM3YE3CAGZJ7KWJBC6J
   Connected Server Name: n3-lon
                     TLS: true (set --tls-detail to see certificates)

....
````

When just doing account info and its tls true add the message as on the last line if false just show `TLS: false`, and if supplied print the chain etc nicely at the bottom.

File: nats/util_tls.go [hunk 2]
```diff
@@ -0,0 +1,96 @@
...
+			})...)
+		}
+		fmt.Print(string(pemData))
+
+		return nil
+	}
+	return nil
+}
+
+var _ nats.Option = enableTLSDebug
```
ripienaar, line 96: probably not needed, it wouldnt compile anyway as you use it as a option in main.go

nats/util_tls.go
```diff
@@ -0,0 +1,96 @@
...
+		// output suggests it was sent by the server.
+		//
+		// To balance these out, we're going to log details of all the chains,
+		// then show just the certs as presented by the server.
+
+		for chain := 0; chain < len(cs.VerifiedChains); chain++ {
+			certCount := len(cs.VerifiedChains[chain])
+			log.Printf("tls-dump-certs: chain %d: certificates: %d", chain+1, certCount)
+			for certI := 0; certI < certCount; certI++ {
+				log.Printf("tls-dump-certs: chain %d: certificate: %d subject: %q",
```
ripienaar, line 64: UX wise these are not great, so when moving into the accounts command I'd create the option with the accounts command as a variable thats in scope and then set the cert chain there in variables.  

Then in the accounts command render it nicely so it looks clean, we can make it look really good then showing various fields like you do here but rendered in line with the general UX of showing information there

ripienaar (changes requested):
Mainly requests about the cosmetics of things

nats/util_tls.go
philpennock: For myself, I like that it explicitly documents here that the function signature is expected to conform to an interface, so don't go trying to change it with new stuff.  :)

ripienaar:
Closing this, feel free to re-open if you later want to work on this again

PR #215, opened 2021-08-16
Adds streaming and request/reply to the bench functionality

jnmoyne:
Adds the following flags to the 'bench' command, introduces JS support and new request/reply mode

example uses:

benchmark core nats publish and subscribe with 10 publishers and subscribers
```
nats bench testsubject --pub 10 --sub 10 --msgs 10000 --size 512
```

benchmark core nats request/reply with queuing
```
nats bench testsubject --sub 4 --reply 
nats bench testsubject --pub 4 --request --msgs 20000
```

benchmark JetStream acknowledged publishes
```
nats bench testsubject --js --syncpub --pub 10  --msgs 10000
```

benchmark JS publish and push consumers at the same time
```
nats bench testsubject --pub 4 --sub 4 --js
```

benchmark just publishing into a JS stream, do not delete the stream so the data in it can be reused by another benchmark (like below)
```
nats bench testsubject --pub 4 --js --nodelete 
```

benchmark JS stream consumption (replay) using a push consumer, keep the stream and it's data around
```
nats bench testsubject --sub 4 --js --nodelete --nopurge
```

benchmark JS stream consumption (replay) using a batched pull consumer, keep the stream and it's data around
```
nats bench testsubject --sub 4 --js --nodelete --nopurge --pull
```

PLEASE NOTE: the default value for --pub was changed from 1 to 0

### DIFF
--- a/nats/bench_command.go
+++ b/nats/bench_command.go
@@ -29,15 +29,31 @@ import (
 )
 
 type benchCmd struct {
-	subject  string
-	numPubs  int
-	numSubs  int
-	numMsg   int
-	msgSize  int
-	csvFile  string
-	progress bool
-	ack      bool
+	subject        string
+	numPubs        int
+	numSubs        int
+	numMsg         int
+	msgSize        int
+	csvFile        string
+	progress       bool
+	syncPub        bool
+	js             bool
+	pubBatch       int
+	jsFile         bool
+	pull           bool
+	pullBatch      int
+	ack            bool
+	replicas       int
+	noPurge        bool
+	noDeleteStream bool
+	maxAckPending  int
 }
+
+const (
+	JS_STREAM_NAME       string = "benchstream"
+	JS_PUSHCONSUMER_NAME string = "pushconsumer"
+	JS_PULLCONSUMER_NAME string = "pullconsumer"
+)
 
 func configureBenchCommand(app *kingpin.Application) {
 	c := &benchCmd{}
@@ -49,13 +65,35 @@ func configureBenchCommand(app *kingpin.Application) {
 	bench.Flag("size", "Size of the test messages").Default("128").IntVar(&c.msgSize)
 	bench.Flag("csv", "Save benchmark data to CSV file").StringVar(&c.csvFile)
 	bench.Flag("progress", "Enable progress bar while publishing").Default("true").BoolVar(&c.progress)
-	bench.Flag("ack", "Waits for acknowledgement on messages using Requests rather than Publish").Default("false").BoolVar(&c.ack)
+	bench.Flag("syncpub", "Synchronously waits for acknowledgement on published messages (using Request in NATS or synchronous Publish in JS)").Default("false").BoolVar(&c.syncPub)
+	bench.Flag("js", "Use JetStream streaming").Default("false").BoolVar(&c.js)
+	bench.Flag("pubbatch", "Sets the batch size for JS asynchronous publishing").Default("100").IntVar(&c.pubBatch)
+	bench.Flag("jsfile", "Persist the stream to file").Default("false").BoolVar(&c.jsFile)
+	bench.Flag("pull", "Uses a JS pull consumer").Default("false").BoolVar(&c.pull)
+	bench.Flag("pullbatch", "Sets the batch size for the JS pull consumer").Default("100").IntVar(&c.pullBatch)
+	bench.Flag("ack", "Acks consumption of messages").Default("false").BoolVar(&c.ack)
+	bench.Flag("replicas", "Number of stream replicas").Default("1").IntVar(&c.replicas)
+	bench.Flag("nopurge", "Do not purge the stream before running").Default("false").BoolVar(&c.noPurge)
+	bench.Flag("nodelete", "Do not delete the stream at the end of the run").Default("false").BoolVar(&c.noDeleteStream)
+	bench.Flag("maxackpending", "Max acks pending for JS consumer").Default("-1").IntVar(&c.maxAckPending)
 
 	cheats["bench"] = `# benchmark JetStream acknowledged publishes
-nats bench --ack --msgs 10000 ORDERS.bench
+nats bench testsubject --syncpub --msgs=10000 ORDERS.bench
 
 # benchmark core nats publish and subscribe with 10 publishers and subscribers
-nats bench --pub 10 --sub 10 --msgs 10000 --size 512
+nats bench testsubject --pub=10 --sub=10 --msgs=10000 --size=512
+
+# benchmark JS publish and push consumers at the same time
+nats bench testsubject --pub=4 --sub=4 --js
+
+# benchmark JS stream purge and async batched publish into the stream
+nats bench testsubject --pub=4 --sub=0 --js --nodelete 
+
+# benchmark JS stream do not purge and get replay from the stream using a push consumer
+nats bench testsubject --pub=0 --sub=4 --js --nodelete --nopurge
+
+# benchmark JS stream do not purge and get replay from the stream using a pull consumer
+nats bench testsubject --pub=0 --sub=4 --js --nodelete --nopurge --pull
 `
 }
 
@@ -64,17 +102,69 @@ func (c *benchCmd) bench(_ *kingpin.ParseContext) error {
 		return fmt.Errorf("number of messages should be greater than 0")
 	}
 
-	log.Printf("Starting benchmark [msgs=%s, msgsize=%s, pubs=%d, subs=%d]", humanize.Comma(int64(c.numMsg)), humanize.IBytes(uint64(c.msgSize)), c.numPubs, c.numSubs)
+	log.Printf("Starting benchmark [msgs=%s, msgsize=%s, pubs=%d, subs=%d, js=%v, jsfile=%v, syncpub=%v, pubbatch=%s, pull=%v, pullbatch=%s, ack=%v, maxackpending=%s, replicas=%d, nopurge=%v, nodelete=%v]", humanize.Comma(int64(c.numMsg)), humanize.IBytes(uint64(c.msgSize)), c.numPubs, c.numSubs, c.js, c.jsFile, c.syncPub, humanize.Comma(int64(c.pubBatch)), c.pull, humanize.Comma(int64(c.pullBatch)), c.ack, humanize.Comma(int64(c.maxAckPending)), c.replicas, c.noPurge, c.noDeleteStream)
 
-	if c.ack && c.progress {
-		log.Printf("Disabling progress bars in request mode")
+	if c.syncPub && c.progress {
+		log.Printf("Disabling progress bars in syncpub mode")
 		c.progress = false
+	}
+
+	if c.js && c.maxAckPending != -1 && c.maxAckPending < c.numMsg && !c.ack && c.numSubs > 0 {
+		log.Printf("WARNING: max acks pending is smaller than the mumber of messages and the subscribers are not set to ack, use --ack or you run the risk of the consumers not receiving all the messages!")
 	}
 
 	bm := bench.NewBenchmark("NATS", c.numSubs, c.numPubs)
 
 	startwg := &sync.WaitGroup{}
 	donewg := &sync.WaitGroup{}
+
+	var js nats.JetStreamContext
+
+	if c.js {
+		// create the stream for the benchmark (and purge it)
+		nc, err := nats.Connect(config.ServerURL(), natsOpts()...)
+		if err != nil {
+			log.Fatalf("nats connection failed: %s", err)
+			return err
+		}
+
+		js, err = nc.JetStream()
+		if err != nil {
+			log.Fatalf("couldn't create jetstream context: %v", err)
+		}
+
+		storageType := func() nats.StorageType {
+			if c.jsFile {
+				return nats.FileStorage
+			} else {
+				return nats.MemoryStorage
+			}
+		}()
+
+		js.AddStream(&nats.StreamConfig{Name: JS_STREAM_NAME, Subjects: []string{c.subject}, Retention: nats.LimitsPolicy, MaxConsumers: -1, MaxMsgs: -1, MaxBytes: -1, Discard: nats.DiscardOld, MaxAge: 9223372036854775807, MaxMsgsPerSubject: -1, MaxMsgSize: -1, Storage: storageType, Replicas: c.replicas, Duplicates: time.Second * 2})
+		if !c.noPurge {
+			log.Printf("Purging the stream")
+			js.PurgeStream(JS_STREAM_NAME)
+		}
+
+		if !c.noDeleteStream {
+			log.Printf("Will delete the stream at the end of the run")
+			defer js.DeleteStream(JS_STREAM_NAME)
+		}
+
+		// create the pull consumer
+		if c.pull && c.numSubs > 0 {
+			js.AddConsumer(JS_STREAM_NAME, &nats.ConsumerConfig{
+				Durable:       JS_PULLCONSUMER_NAME,
+				DeliverPolicy: nats.DeliverAllPolicy,
+				AckPolicy:     nats.AckExplicitPolicy,
+				ReplayPolicy:  nats.ReplayInstantPolicy,
+				MaxAckPending: c.maxAckPending,
+			})
+			defer js.DeleteConsumer(JS_STREAM_NAME, JS_PULLCONSUMER_NAME)
+		}
+	}
+	subCounts := bench.MsgsPerClient(c.numMsg, c.numSubs)
 
 	for i := 0; i < c.numSubs; i++ {
 		nc, err := nats.Connect(config.ServerURL(), natsOpts()...)
@@ -86,11 +176,20 @@ func (c *benchCmd) bench(_ *kingpin.ParseContext) error {
 		startwg.Add(1)
 		donewg.Add(1)
 
-		go c.runSubscriber(bm, nc, startwg, donewg)
+		numMsg := func() int {
+			if c.pull {
+				return subCounts[i]
+			} else {
+				return c.numMsg
+			}
+		}()
+
+		go c.runSubscriber(bm, nc, startwg, donewg, numMsg)
 	}
 	startwg.Wait()
 
 	pubCounts := bench.MsgsPerClient(c.numMsg, c.numPubs)
+
 	for i := 0; i < c.numPubs; i++ {
 		nc, err := nats.Connect(config.ServerURL(), natsOpts()...)
 		if err != nil {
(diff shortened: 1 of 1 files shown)


nats/bench_command.go
```diff
@@ -49,10 +62,17 @@ func configureBenchCommand(app *kingpin.Application) {
 	bench.Flag("size", "Size of the test messages").Default("128").IntVar(&c.msgSize)
 	bench.Flag("csv", "Save benchmark data to CSV file").StringVar(&c.csvFile)
 	bench.Flag("progress", "Enable progress bar while publishing").Default("true").BoolVar(&c.progress)
-	bench.Flag("ack", "Waits for acknowledgement on messages using Requests rather than Publish").Default("false").BoolVar(&c.ack)
+	bench.Flag("request", "Waits for acknowledgement on messages using Requests rather than Publish").Default("false").BoolVar(&c.request)
+	bench.Flag("streaming", "Use JetStream streaming").Default("false").BoolVar(&c.js)
```
wallyqs, line 66: maybe `-js` like in the server?

nats/bench_command.go
```diff
@@ -181,29 +270,69 @@ func (c *benchCmd) runPublisher(bm *bench.Benchmark, nc *nats.Conn, startwg *syn
...
+		sub, _ = nc.Subscribe(c.subject, mh)
+	}
 
 	sub.SetPendingLimits(-1, -1)
 	nc.Flush()
 	startwg.Done()
 
+	if c.js && c.pull {
+		for i := 0; i < numMsg; i++ {
+			msg, err := sub.NextMsg(time.Second * 3)
```
wallyqs, line 318: For a pull consumer here have to call `Fetch(n)` where n is the batch number, could be 1 for a single message.

nats/bench_command.go
```diff
@@ -181,29 +270,70 @@ func (c *benchCmd) runPublisher(bm *bench.Benchmark, nc *nats.Conn, startwg *syn
...
+		sub, _ = nc.Subscribe(c.subject, mh)
+	}
 
 	sub.SetPendingLimits(-1, -1)
 	nc.Flush()
 	startwg.Done()
 
+	if c.js && c.pull {
+		for i := 0; i < numMsg; i++ {
+			msgs, err := sub.Fetch(1)
```
wallyqs, line 317: you can change the timeout here with `natx.MaxWait()` if needed

jnmoyne: Yes, will add batch size functionality

nats/bench_command.go
jnmoyne: That's how I had it initially, but then thought that the new messaging is that JS is the new persistence layer of NATS and that now streaming is implemented with JS ... but there are also other things (e.g. kv store) on top of JS as well as streaming, so I changed it to --streaming (and the help says 'Use JetStream streaming'). But open to changing it back to --js if using --streaming feels wrong to you.

wallyqs: Makes sense though I'd leave it as `-js` tot be consistent with the server for now since nats streaming has a couple of years still to be deprecated.

nats/bench_command.go
jnmoyne: --pullbatch added

nats/bench_command.go
jnmoyne: reverted back to --js

derekcollison:
Possible to squash down for easier review?

vishaltripathi24:
Awaiting this script, what's the eta for it to release?, Thanks

nats/bench_command.go
```diff
@@ -49,13 +65,35 @@ func configureBenchCommand(app *kingpin.Application) {
...
+	bench.Flag("syncpub", "Synchronously waits for acknowledgement on published messages (using Request in NATS or synchronous Publish in JS)").Default("false").BoolVar(&c.syncPub)
+	bench.Flag("js", "Use JetStream streaming").Default("false").BoolVar(&c.js)
+	bench.Flag("pubbatch", "Sets the batch size for JS asynchronous publishing").Default("100").IntVar(&c.pubBatch)
+	bench.Flag("jsfile", "Persist the stream to file").Default("false").BoolVar(&c.jsFile)
+	bench.Flag("pull", "Uses a JS pull consumer").Default("false").BoolVar(&c.pull)
+	bench.Flag("pullbatch", "Sets the batch size for the JS pull consumer").Default("100").IntVar(&c.pullBatch)
+	bench.Flag("ack", "Acks consumption of messages").Default("false").BoolVar(&c.ack)
+	bench.Flag("replicas", "Number of stream replicas").Default("1").IntVar(&c.replicas)
+	bench.Flag("nopurge", "Do not purge the stream before running").Default("false").BoolVar(&c.noPurge)
+	bench.Flag("nodelete", "Do not delete the stream at the end of the run").Default("false").BoolVar(&c.noDeleteStream)
```
ColinSullivan1, line 77: I like how the tools will clean up after itself.  +1

nats/bench_command.go
```diff
@@ -64,18 +102,70 @@ func (c *benchCmd) bench(_ *kingpin.ParseContext) error {
...
 
-	if c.ack && c.progress {
-		log.Printf("Disabling progress bars in request mode")
+	if c.syncPub && c.progress {
+		log.Printf("Disabling progress bars in syncpub mode")
 		c.progress = false
 	}
 
+	if c.js && c.maxAckPending != -1 && c.maxAckPending < c.numMsg && !c.ack && c.numSubs > 0 {
+		log.Printf("WARNING: max acks pending is smaller than the mumber of messages and the subscribers are not set to ack, use --ack or you run the risk of the consumers not receiving all the messages!")
```
ColinSullivan1, line 113: a nitpick - other log lines do not have a `!`

nats/bench_command.go
```diff
@@ -49,13 +65,35 @@ func configureBenchCommand(app *kingpin.Application) {
 	bench.Flag("size", "Size of the test messages").Default("128").IntVar(&c.msgSize)
 	bench.Flag("csv", "Save benchmark data to CSV file").StringVar(&c.csvFile)
 	bench.Flag("progress", "Enable progress bar while publishing").Default("true").BoolVar(&c.progress)
-	bench.Flag("ack", "Waits for acknowledgement on messages using Requests rather than Publish").Default("false").BoolVar(&c.ack)
+	bench.Flag("syncpub", "Synchronously waits for acknowledgement on published messages (using Request in NATS or synchronous Publish in JS)").Default("false").BoolVar(&c.syncPub)
+	bench.Flag("js", "Use JetStream streaming").Default("false").BoolVar(&c.js)
+	bench.Flag("pubbatch", "Sets the batch size for JS asynchronous publishing").Default("100").IntVar(&c.pubBatch)
+	bench.Flag("jsfile", "Persist the stream to file").Default("false").BoolVar(&c.jsFile)
```
ColinSullivan1, line 71: to future-proof might want to have a general type just in case we add additional storage types.

ColinSullivan1:
Overall looks ok, just nitpicks.  Should also fix the PR description.  The log messages start with a mix of lowercase and uppercase, should standardize - it looks like the rest of the messages start with lowercase.

nats/bench_command.go
jnmoyne: Done

nats/bench_command.go
jnmoyne: Done

nats/bench_command.go
ripienaar: Boolean flags like `--purge` would automatically support `--no-purge` as a side effect, the help doesnt show them clearly so for these lets make them `--no-delete` and `--no-purge` for consistency. Also `--no-progress` and `--no-queue`.

Relying on the `--purge` and supporting `--no-purge` as side effect also avoid all these stupid go negative booleans, we have statements like `if !noPurge {...}` which is one of the worst things in go forcing people to do write double negatives all the time.  So had these variables been `purge` and default to `"true"` code all reads easier.  You can still have `--no-purge` if you like but call the variable `purge` and defaults to `"true"` in the CLI to avoid the double negatives

nats/bench_command.go
```diff
@@ -29,33 +29,82 @@ import (
...
+nats bench testsubject --pub 1 --sub 1 --msgs 10000 --noqueue
+
+# benchmark core nats request/reply with queuing
+nats bench testsubject --sub 4 --reply
+nats bench testsubject --pub 4 --request --msgs 20000
+
+# benchmark JetStream acknowledged publishes
+nats bench testsubject --js --syncpub --pub 10  --msgs 10000
+
+
```
ripienaar, line 96: drop the extra blank line here

nats/bench_command.go
```diff
@@ -64,37 +113,123 @@ func (c *benchCmd) bench(_ *kingpin.ParseContext) error {
...
 	}
 
-	log.Printf("Starting benchmark [msgs=%s, msgsize=%s, pubs=%d, subs=%d]", humanize.Comma(int64(c.numMsg)), humanize.IBytes(uint64(c.msgSize)), c.numPubs, c.numSubs)
+	log.Printf("Starting benchmark [msgs=%s, msgsize=%s, pubs=%d, subs=%d, js=%v, storage=%s, syncpub=%v, pubbatch=%s, pull=%v, pullbatch=%s, request=%v, reply=%v, noqueue=%v, maxackpending=%s, replicas=%d, nopurge=%v, nodelete=%v]", humanize.Comma(int64(c.numMsg)), humanize.IBytes(uint64(c.msgSize)), c.numPubs, c.numSubs, c.js, c.storage, c.syncPub, humanize.Comma(int64(c.pubBatch)), c.pull, humanize.Comma(int64(c.pullBatch)), c.request, c.reply, c.noQueueGroup, humanize.Comma(int64(c.maxAckPending)), c.replicas, c.noPurge, c.noDeleteStream)
 
-	if c.ack && c.progress {
-		log.Printf("Disabling progress bars in request mode")
-		c.progress = false
+	if c.numPubs == 0 && c.numSubs == 0 {
+		log.Fatalf("You must have at least one publisher or at least one subscriber... try adding --pub 1 and/or --sub 1 to the arguments")
```
ripienaar, line 119: so wouldnt the previous default of 1 on pub make more sense? That way the default happy path is to public, if you really want to set sub only you can disable that. The tool should function with as few as possible flags

nats/bench_command.go
```diff
@@ -64,37 +113,123 @@ func (c *benchCmd) bench(_ *kingpin.ParseContext) error {
...
+	}
+
+	if c.js && c.maxAckPending != -1 && c.maxAckPending < c.numMsg && c.numSubs > 0 {
+		log.Printf("WARNING: max acks pending is smaller than the mumber of messages and the subscribers are not set to ack, use --ack or you run the risk of the consumers not receiving all the messages")
```
ripienaar, line 123: For benchmarking this situation seems fatal not a warning to me, we should help people perform correct benchmarks by either failing or adjustin their settings with a warning, we shouldnt let potentially broken benchmarks be done

nats/bench_command.go
```diff
@@ -64,37 +113,123 @@ func (c *benchCmd) bench(_ *kingpin.ParseContext) error {
...
 
 	bm := bench.NewBenchmark("NATS", c.numSubs, c.numPubs)
 
 	startwg := &sync.WaitGroup{}
 	donewg := &sync.WaitGroup{}
 
+	var js nats.JetStreamContext
+
+	if (c.request || c.reply) && c.js {
+		log.Fatalf("Request/Reply mode is not applicable to streams")
```
ripienaar, line 134: `...applicable to JetStream benchmarking` we use the word `streams` in too many places

nats/bench_command.go
```diff
@@ -64,37 +113,123 @@ func (c *benchCmd) bench(_ *kingpin.ParseContext) error {
...
+		}
+
+		js, err = nc.JetStream()
+		if err != nil {
+			log.Fatalf("Couldn't create jetstream context: %v", err)
+		}
+
+		// First delete any prior stream unless no delete
+		if !c.noDeleteStream {
+			log.Printf("Deleting any existing stream")
```
ripienaar, line 162: its worth mentioning what the stream is called in the message

nats/bench_command.go
```diff
@@ -64,37 +113,123 @@ func (c *benchCmd) bench(_ *kingpin.ParseContext) error {
...
+			js.DeleteStream(JS_STREAM_NAME)
```
ripienaar, line 163: handle error

nats/bench_command.go
```diff
@@ -64,37 +113,123 @@ func (c *benchCmd) bench(_ *kingpin.ParseContext) error {
...
+				return nats.MemoryStorage
+			default:
+				{
+					log.Printf("Unknown storage type %s, using memory", c.storage)
+					return nats.MemoryStorage
+				}
+			}
+		}()
+
+		js.AddStream(&nats.StreamConfig{Name: JS_STREAM_NAME, Subjects: []string{c.subject}, Retention: nats.LimitsPolicy, MaxConsumers: -1, MaxMsgs: -1, MaxBytes: -1, Discard: nats.DiscardOld, MaxAge: 9223372036854775807, MaxMsgsPerSubject: -1, MaxMsgSize: -1, Storage: storageType, Replicas: c.replicas, Duplicates: time.Second * 2})
```
ripienaar, line 180: surely all these -1 and maxage of 9223372036854775807 isnt required, the js context would do the right thing?

nats/bench_command.go
```diff
@@ -64,37 +113,123 @@ func (c *benchCmd) bench(_ *kingpin.ParseContext) error {
...
+		if !c.noPurge {
```
ripienaar, line 181: avoid double negative here

nats/bench_command.go
```diff
@@ -64,37 +113,123 @@ func (c *benchCmd) bench(_ *kingpin.ParseContext) error {
...
+			log.Printf("Purging the stream")
+			js.PurgeStream(JS_STREAM_NAME)
+		}
+
+		if !c.noDeleteStream {
```
ripienaar, line 186: avoid double negative here

nats/bench_command.go
```diff
@@ -64,37 +113,123 @@ func (c *benchCmd) bench(_ *kingpin.ParseContext) error {
...
+			defer js.DeleteConsumer(JS_STREAM_NAME, JS_PULLCONSUMER_NAME)
+		}
+	}
+	subCounts := bench.MsgsPerClient(c.numMsg, c.numSubs)
+
 	for i := 0; i < c.numSubs; i++ {
 		nc, err := nats.Connect(config.ServerURL(), natsOpts()...)
 		if err != nil {
-			return fmt.Errorf("nats connection %d failed: %s", i, err)
+			return fmt.Errorf("NATS connection %d failed: %s", i, err)
```
ripienaar, line 208: return `nats connection %d failed` as before this will probably fail linting, idiomatic go is to return lower case first word in erorrs.

nats/bench_command.go
```diff
@@ -64,37 +113,123 @@ func (c *benchCmd) bench(_ *kingpin.ParseContext) error {
...
+
+		js.AddStream(&nats.StreamConfig{Name: JS_STREAM_NAME, Subjects: []string{c.subject}, Retention: nats.LimitsPolicy, MaxConsumers: -1, MaxMsgs: -1, MaxBytes: -1, Discard: nats.DiscardOld, MaxAge: 9223372036854775807, MaxMsgsPerSubject: -1, MaxMsgSize: -1, Storage: storageType, Replicas: c.replicas, Duplicates: time.Second * 2})
+		if !c.noPurge {
+			log.Printf("Purging the stream")
+			js.PurgeStream(JS_STREAM_NAME)
+		}
+
+		if !c.noDeleteStream {
+			log.Printf("Will delete the stream at the end of the run")
+			defer js.DeleteStream(JS_STREAM_NAME)
```
ripienaar, line 188: should probably handle the error and warn if it wasnt deleted

nats/bench_command.go
```diff
@@ -64,37 +113,123 @@ func (c *benchCmd) bench(_ *kingpin.ParseContext) error {
...
 	}
 	startwg.Wait()
 
 	pubCounts := bench.MsgsPerClient(c.numMsg, c.numPubs)
+
 	for i := 0; i < c.numPubs; i++ {
 		nc, err := nats.Connect(config.ServerURL(), natsOpts()...)
 		if err != nil {
-			return fmt.Errorf("nats connection %d failed: %s", i, err)
+			return fmt.Errorf("NATS connection %d failed: %s", i, err)
```
ripienaar, line 232: keep lower case error

nats/bench_command.go
```diff
@@ -104,7 +239,7 @@ func (c *benchCmd) bench(_ *kingpin.ParseContext) error {
 		go c.runPublisher(bm, nc, startwg, donewg, pubCounts[i])
 	}
 
-	if c.progress {
+	if !c.noProgress {
```
ripienaar, line 242: avoid double negative

nats/bench_command.go
```diff
@@ -113,7 +248,7 @@ func (c *benchCmd) bench(_ *kingpin.ParseContext) error {
 
 	bm.Close()
 
-	if c.progress {
+	if !c.noProgress {
```
ripienaar, line 251: avoid double negative

nats/bench_command.go
```diff
@@ -133,7 +268,7 @@ func (c *benchCmd) runPublisher(bm *bench.Benchmark, nc *nats.Conn, startwg *syn
 	startwg.Done()
 
 	var progress *uiprogress.Bar
-	if c.progress {
+	if !c.noProgress {
```
ripienaar, line 271: avoid double negative

nats/bench_command.go
```diff
@@ -147,30 +282,68 @@ func (c *benchCmd) runPublisher(bm *bench.Benchmark, nc *nats.Conn, startwg *syn
...
+
+			if !c.request {
+				err = nc.Publish(c.subject, msg)
+				if err != nil {
+					log.Fatalf("Publish error: %s", err)
+				}
+			} else {
+				m, err = nc.Request(c.subject, msg, time.Second)
+				if err != nil {
+					log.Fatalf("Request error: %s", err)
```
ripienaar, line 306: should a single request failing really terminate the entire benchmark?

nats/bench_command.go
```diff
@@ -147,30 +282,68 @@ func (c *benchCmd) runPublisher(bm *bench.Benchmark, nc *nats.Conn, startwg *syn
...
+		} else {
+			for i := 0; i < numMsg; i++ {
+				if progress != nil {
+					progress.Incr()
+				}
+				_, err = js.Publish(c.subject, msg)
+				if err != nil {
+					log.Fatalf("Publish error: %s", err)
+				}
+			}
```
ripienaar, line 346: this method has turned into a bit of a monster with several sets of if js / else branching, might be easier to grok if there was `benchmarkJS` and `benchmarkCore` or similar functions

ripienaar (changes requested):
looks great, mostly cosmetic changes

nats/bench_command.go
jnmoyne: I took that default out because I really didn't like when you would use

`nats bench testsubject --sub 1 ...` 

It would then end up creating a subscriber _and_ a publisher (which can definitely surprise you)

So you end up always having to specify `--pub=0` if you do not want to run a publisher. And in this new version of nats bench most of the time you will run it with either no subscriber(s) or no publisher(s) (while before since no JS mode of operation you would maybe use both a publisher and a subscriber more often - which BTW I don't consider a very realistic scenario: in 'real life' your application usually don't just subscribe to their own publications)

I agree that you want to keep the number of required flags as few as you can, but in this case having that pub default to 1 and sub default to 1, besides sometimes surprising you, means there are many times you need to add a `--pub 0` argument (so actually it means the old way would cause you to have to add more flags than now, most of the time since now you most of the time would not run both a publisher and a subscriber flags at the same time).

I really don't want to change what I did, once you get used to it a bit it (to me at least) makes more sense to have both --pub and --sub default to 0 (than having just one of them default to 0 and the other to 1 like it was)

nats/bench_command.go
```diff
@@ -76,6 +96,54 @@ func (c *benchCmd) bench(_ *kingpin.ParseContext) error {
...
 	donewg := &sync.WaitGroup{}
 
+	var js nats.JetStreamContext
+
+	if c.js {
+		// create the stream for the benchmark (and purge it)
+		nc, err := nats.Connect(config.ServerURL(), natsOpts()...)
+		if err != nil {
+			log.Fatalf("nats connection failed: %s", err)
+			return err
```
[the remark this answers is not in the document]
jnmoyne: return removed (those few lines I originally copied some somewhere else, hence the return was there)

nats/bench_command.go
jnmoyne: done

nats/bench_command.go
ripienaar: OK

nats/bench_command.go
jnmoyne: As I mentioned elsewhere I changed to using --noFoo format because many people may not know about the --no-foo side effect to unset the flag (and because doing something like --foo=false doesn't work). I don't mind changing back to --foo but will then add a reminder of how to unset a flag set to true by default in the help.

ripienaar: Please explain the difference between:

```
 	bench.Flag("nodelete", "Do not delete the stream at the end of the run").Default("false").BoolVar(&c.noDeleteStream) 
```

and


```
 	bench.Flag("no-delete", "Do not delete the stream at the end of the run").Default("true").BoolVar(&c.deleteStream) 
```

ripienaar: (sorry, that was a bad example let me see where I did this correctly)

ripienaar: There's a nasty way to work around this - so for now lets just use `bench.Flag("no-delete".....)` and not `bench.Flag("nodelete",....)` thus keeping the `--no-*` pattern consistent but keep the variable names as is.

nats/bench_command.go
jnmoyne: done

nats/bench_command.go
jnmoyne: done

nats/bench_command.go
jnmoyne: done

nats/bench_command.go
jnmoyne: it passes lint but I changed it back (I had capitalized it because I thought it looked better, didn't know about that idiomatic Go aspect), changed back to all lower case now

nats/bench_command.go
jnmoyne: done

nats/bench_command.go
jnmoyne: Done

nats/bench_command.go
jnmoyne: Done like that by request from Colin (thought being that if the subscriber is expecting a specific number of messages it would then hang for ever (at least as a subscriber) if the publisher is unable to publish one or more of it's messages

nats/bench_command.go
jnmoyne: Wanted to make sure that those values (which should be large enough for any kind of benchmark scenario) were explicitly set in the source code.

nats/bench_command.go
jnmoyne: sure...

nats/bench_command.go
jnmoyne: done

ripienaar (approved):
LGTM

PR #225, opened 2021-08-23
Update readme with new nats bench examples

jnmoyne:
Updated the readme file as it had now outdated nats bench examples


ripienaar (changes requested):
LGTM, though the changes in the list under features isnt needed

ripienaar (approved):

PR #231, opened 2021-08-27
Added --pubtimeout to control the timeout for JS asynch publishing

jnmoyne:
Added --pubtimeout to control the timeout for JS asynch publishing and changed the default to a more sane 30s. Re-worked and improved progress bars, which now work for subscribing as well as publishing.


ripienaar (approved):
LGTM

PR #238, opened 2021-08-31
Added bench --stream flag to specify the stream name to use for the benchmark

jnmoyne:


ripienaar:
Bit worried someone will point this at an existing stream thinking they want to test their exact co fig and then we delete it.

jnmoyne:
So am I, part of the reason it wasn't exposed initially, and why I added a warning to the help (warning the stream will be deleted/purged unless --no-delete/--no-purge are set) for this new argument.

Maybe the better choice is to change --no-delete to a --delete so the default behavior is to not try to delete the stream before/after the run (and by extension --no-purge to be a --purge. That way there's less chance they delete an existing stream by default because they forgot to add --no-delete or --no-purge.

jnmoyne:
We could, but given how in practice when you are doing JS benchmarks since you need to measure publication and subscription rates separately and end up almost always having to set --no-delete and --no-purge IMHO I'd rather do the change of those flags rather than error out if you specify a stream name unless you include --no-delete: i.e. equivalent to breaking the --no-delete argument if you also specify --stream

nats/bench_command.go
```diff
@@ -159,7 +159,7 @@ func (c *benchCmd) bench(_ *kingpin.ParseContext) error {
 		}
 
 		// First delete any prior stream unless no delete
-		if !c.noDeleteStream {
+		if c.deleteStream {
```
ripienaar, line 162: so part of deleting it is that it's safe to do this

https://github.com/nats-io/natscli/blob/57a26979704d591774f0deeaba412e120f2936c8/nats/bench_command.go#L184-L185

What we cant have is if someone gives a stream name that we then go and edit the properties of that stream, if they give a stream name and it exist, we either have to somehow adapt to it and bench against it or something? 

I don't know, to be honest the ability to take arbitrary stream names seems like a trap to me.

When we create the consumer, there's no filter subjects - so again on a pre-existing stream this will go bad. So we'd need to ensure we filter only on the given c.subject

I am more and more unconvinced about the ability to operate on anything but a clean environment ie. stream must not exist before we start.  We should never `make sure the existing one has the same attributes`.

ripienaar: Sure for consumption obviously the stream must exist. But what about publishing is it not a safe approach to say the stream must not be there? Anyway as you clarify the client will fail for differences in config so it’s probsbly fine. 

Sounds good we can take either approach with delete or not etc - I am not trying to dictate the design I just want to be sure it’s safe for users if they point it at a stream with prod data that’s all

jnmoyne: So I went with removing the --delete functionality altogether (and improving the error messages for the cases where the user may need to manually delete the stream first). Telling them to use `nats stream rm` (rather than deleting directly) also means they get prompted "are you sure?" before the delete actually happens.

ripienaar:
LGTM!

PR #241, opened 2021-09-02
fix nats connect call

jnmoyne:
nats connect needs url,natsOpts()... in order for the connection with –tlsca and –creds to work (was mistakenly removed by some earlier commit thinking it was un-needed)


ripienaar (approved):
LGTM

PR #249, opened 2021-09-23
Improve logging and progress bar

jnmoyne:
Also remove --no-queue which was only for --reply since it doesn't make sense in typical request/reply use cases to do not have repliers in a queue group (as it would lead to duplicate processing) why offer it as a kind of benchmark.


ripienaar (approved):
LGTM

PR #262, opened 2021-10-11
nats.go kv and obj store

ripienaar:
needs kv.Status() implemented


ripienaar:
* Needs https://github.com/nats-io/nats.go/pull/844 to have correct delete behavior
* Needs https://github.com/nats-io/nats.go/pull/843 to avoid spelunking into the stream api

PR #272, opened 2021-10-18
Updated client to fix object store get over long rtt

derekcollison:
Signed-off-by: Derek Collison <derek@nats.io>


ripienaar:
Weird extra dependencies that I don’t think we need here

PR #315, opened 2021-12-11
adding --count to reply and --reply to request

### LINKED ISSUE #314: no way to set the reply subject in nats request
Something where I could either se the entire subject.
Or better yet, subject with template "customprefix.{{inbox}}"

```
nats request -h
usage: nats request [<flags>] <subject> [<body>]

Generic data request utility

Body and Header values of the messages may use Go templates to create unique messages.

  nats pub test --count 10 "Message {{Count}} @ {{Time}}"

Multiple messages with random strings between 10 and 100 long:

  nats pub test --count 10 "Message {{Count}}: {{ Random 10 100 }}"

Available template functions are:

  Count            the message number
  TimeStamp        RFC3339 format current time
  Unix             seconds since 1970 in UTC
  UnixNano         nano seconds since 1970 in UTC
  Time             the current time
  ID               an unique ID
  Random(min, max) random string at least min long, at most max

Flags:
  -h, --help                    Show context-sensitive help (also try --help-long and --help-man).
      --version                 Show application version.
  -s, --server=NATS_URL         NATS server urls
      --user=NATS_USER          Username or Token
      --password=NATS_PASSWORD  Password
      --creds=NATS_CREDS        User credentials
      --nkey=NATS_NKEY          User NKEY
      --tlscert=NATS_CERT       TLS public certificate
      --tlskey=NATS_KEY         TLS private key
      --tlsca=NATS_CA           TLS certificate authority chain
      --timeout=NATS_TIMEOUT    Time to wait on responses from NATS
      --js-api-prefix=PREFIX    Subject prefix for access to JetStream API
      --js-event-prefix=PREFIX  Subject prefix for access to JetStream Advisories
      --js-domain=DOMAIN        JetStream domain to access
      --context=CONTEXT         Configuration context
      --trace                   Trace API interactions
  -r, --raw                     Show just the output received
  -H, --header=HEADER ...       Adds headers to the message
      --count=1                 Publish multiple messages

Args:
  <subject>  Subject to subscribe to
  [<body>]   Message body
```

matthiashanel:
--count limits the number of messages to reply to before exiting
--reply allows to specify the reply subject

fixes #314

requesting with a different subject (instead of a plain inbox) is a pattern nats allows for. 
Which is why natscli should support it natively, just to make demos easier.
When I experimented with this, the go api, always overwrote the reply subject with an inbox.
Which is why code to avoid the request was added too.

Signed-off-by: Matthias Hanel <mh@synadia.com>

### DIFF
--- a/cli/pub_command.go
+++ b/cli/pub_command.go
@@ -88,6 +88,7 @@ nats request destination.subject "hello world" -H "Content-type:text/plain" --ra
 	req.Flag("raw", "Show just the output received").Short('r').Default("false").BoolVar(&c.raw)
 	req.Flag("header", "Adds headers to the message").Short('H').StringsVar(&c.hdrs)
 	req.Flag("count", "Publish multiple messages").Default("1").IntVar(&c.cnt)
+	req.Flag("reply", "Sets a custom reply to subject").StringVar(&c.replyTo)
 }
 
 func init() {
@@ -119,7 +120,23 @@ func (c *pubCmd) doReq(nc *nats.Conn, progress *uiprogress.Bar) error {
 			return err
 		}
 
-		m, err := nc.RequestMsg(msg, opts.Timeout)
+		var m *nats.Msg
+		if c.replyTo == "" {
+			m, err = nc.RequestMsg(msg, opts.Timeout)
+			if err != nil {
+				return err
+			}
+		} else {
+			sub, e := nc.SubscribeSync(c.replyTo)
+			if e != nil {
+				return e
+			}
+			if e := nc.PublishMsg(msg); e != nil {
+				return e
+			}
+			m, err = sub.NextMsg(opts.Timeout)
+		}
+
 		if err != nil {
 			return err
 		}
--- a/cli/pub_command.go
+++ b/cli/pub_command.go
@@ -127,18 +127,18 @@ func (c *pubCmd) doReq(nc *nats.Conn, progress *uiprogress.Bar) error {
 				return err
 			}
 		} else {
-			sub, e := nc.SubscribeSync(c.replyTo)
-			if e != nil {
-				return e
+			sub, err := nc.SubscribeSync(c.replyTo)
+			if err != nil {
+				return err
 			}
-			if e := nc.PublishMsg(msg); e != nil {
-				return e
+			err = nc.PublishMsg(msg)
+			if  err != nil {
+				return err
 			}
 			m, err = sub.NextMsg(opts.Timeout)
-		}
-
-		if err != nil {
-			return err
+			if err != nil {
+				return err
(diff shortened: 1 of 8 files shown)


cli/pub_command.go
```diff
@@ -119,7 +120,23 @@ func (c *pubCmd) doReq(nc *nats.Conn, progress *uiprogress.Bar) error {
...
 
-		m, err := nc.RequestMsg(msg, opts.Timeout)
+		var m *nats.Msg
+		if c.replyTo == "" {
+			m, err = nc.RequestMsg(msg, opts.Timeout)
+			if err != nil {
+				return err
+			}
+		} else {
+			sub, e := nc.SubscribeSync(c.replyTo)
```
ripienaar, line 130: `err` for error please, also below

cli/pub_command.go
```diff
@@ -119,7 +120,23 @@ func (c *pubCmd) doReq(nc *nats.Conn, progress *uiprogress.Bar) error {
...
+			if e != nil {
+				return e
+			}
+			if e := nc.PublishMsg(msg); e != nil {
```
ripienaar, line 134: split this ie. 

```
err = nc.Publish....
if err != nil { }
```

Thats consistent with the style already in place here

cli/pub_command.go
```diff
@@ -119,7 +120,23 @@ func (c *pubCmd) doReq(nc *nats.Conn, progress *uiprogress.Bar) error {
...
+				return e
+			}
+			m, err = sub.NextMsg(opts.Timeout)
+		}
+
 		if err != nil {
```
ripienaar, line 140: move the error check up so it checks after the err is set, easy mistake to make later to drop this or insert stuff in between

ripienaar (changes requested):
generally ok, some inconsisten style stuff.

Since we're essentially implementing our own request() here it seems it would be simpler to alwas use the custom request logic and just set reply subject in prepareMsg accordingly - either user supplied or a inbox

ripienaar:
A better option might be to make the reply prefix configurable in context and in main settings, what do you think?

cli/pub_command.go
matthiashanel: fixed

cli/pub_command.go
matthiashanel: fixed

cli/pub_command.go
matthiashanel: fixed

matthiashanel:
@ripienaar making use of custom inbox prefix for request now

ripienaar (approved):
LGTM, maybe squash as there's now a bunch of irrelevant history

PR #316, opened 2021-12-11
Add support for the sleep duration in requests

bruth:
Signed-off-by: Byron Ruth <b@devel.io>


bruth:
@ripienaar The issue appears to be when `c.req` is set (via the `wait` option). It is delegated [here](https://github.com/nats-io/natscli/blob/main/cli/pub_command.go#L190-L192) and then the [sleep check](https://github.com/nats-io/natscli/blob/main/cli/pub_command.go#L216-L218) is not applied.

ripienaar:
good stuff, thanks

Issue #317, opened 2021-12-15
Nats CLI does not read from stdin when not run from a terminal / TTY


argl:
It is not possible to run the nats cli in a non-tty shell, for example as a part of a shell script that is run as a docker container entrypoint because there is no tty associated with the process.

Example: given the script
```
#!/usr/bin/env bash
nats pub "some.subject" <payload.json
```

This runs fine when called from a terminal, but fails to read the payload when run headless (i.e. as a docker container entrypoint).

I believe the code in question is in `cli/pub_command.go` ine 169ff:

```
	if c.body == "!nil!" && terminal.IsTerminal(int(os.Stdout.Fd())) { // <-- why the check for a terminal? (and why stdout)
		log.Println("Reading payload from STDIN")
		body, err := ioutil.ReadAll(os.Stdin)
		if err != nil {
			return err
		}
		c.body = string(body)
	}
```

I am not too familiar with the world of ttys in that context, but it seems to be awkward that a command line utility is unusable in a headless context. But I may have missed something important here, so feel free to dump this issue if appropriate, I am sorry for any noise in advance.

ripienaar:
We could add a flag to force reading from stdin, its always a trade off between unintionally reading stdin leading to "its hung and not doing anything" vs not reading it enough.

argl:
Something like that would be very cool. 

FWIW, I built a version with the `isTerminal` condition removed and it works like a charm for my use case now. 

IIUC, this change would prevent the tool from working correctly when intentionally sending a `nil` payload by not passing a literal payload or a readable stdin, right? Other tools use the `-` moniker to force reading from stdin, but I guess this is not backwards compatible.

argl:
It took me an embarrassingly long time to get to the bottom of this, I think having a dedicated command line arg for this would have pointed me into the right direction.

I am happy to prepare a PR for that, is this something considered an easy task (go is not a primary language for me)? Anything to look out for?

ripienaar:
It should be fairly easy to add, add a boolean variable and a option near the top of the file and just check for that around the same place as the terminal check

argl:
OK, I will run with that.

argl:
See https://github.com/nats-io/natscli/pull/318

PR #318, opened 2021-12-15
Added --force-stdin flag to pub command. Enables to read from stdin, …

argl:
…even in a non-tty environment.


argl:
This is a solution for https://github.com/nats-io/natscli/issues/317

ripienaar:
Thank you, well done!

PR #320, opened 2021-12-27
add rmAction to call DeleteObjectStore

miraculli:
I´ve seen that there is no way to remove a ObjectStore with natscli so added a Function, Commands and Action like it is handled in `kv_command.go `

**BUT**: `rm` (and `del`) was used different to the `kv` command:
in kv context `rm` removes the bucket and `del` deletes the key. 
With this patch I also align this to the `object` command: `rm` removes the ObejectStore und `del` deletes a object.

Maybe some docs need to be updated too.
If this change ist considered wrong I can rework this merge request....

### DIFF
--- a/cli/object_command.go
+++ b/cli/object_command.go
@@ -75,10 +75,10 @@ NOTE: This is an experimental feature.
 	put.Flag("no-progress", "Disables progress bars").Default("false").BoolVar(&c.noProgress)
 	put.Flag("force", "Act without confirmation").Short('f').BoolVar(&c.force)
 
-	rm := obj.Command("rm", "Removes a file from the store").Action(c.rmAction)
-	rm.Arg("bucket", "The bucket to act on").Required().StringVar(&c.bucket)
-	rm.Arg("file", "The file to retrieve").Required().StringVar(&c.file)
-	rm.Flag("force", "Act without confirmation").Short('f').BoolVar(&c.force)
+	del := obj.Command("del", "Removes a file from the store").Action(c.delAction)
+	del.Arg("bucket", "The bucket to act on").Required().StringVar(&c.bucket)
+	del.Arg("file", "The file to retrieve").Required().StringVar(&c.file)
+	del.Flag("force", "Act without confirmation").Short('f').BoolVar(&c.force)
 
 	get := obj.Command("get", "Retrieves a file from the store").Action(c.getAction)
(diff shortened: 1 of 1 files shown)


ripienaar:
Thank you.

I have some refinement to do here, I think its comfusing that there is `kv del bucket key` and `kv rm bucket` when this could just as easily be `kv rm bucket` and `kv rm bucket key` and then just have one commend.  I find remembering which is rm and which is del just too hard.

miraculli:
Yes, I thought about that too. But imho the mix of windows (`del`) and unix (`rm`) style is what confuses me most. The aws-cli tool has also special commands for buckets and objects.

Doing stuff with objects:
- cp
- mv
- rm

Doing stuff with bucket:
- mb  (make bucket)
- rb (remove bucket)
- ls
- sync

The [min.io client](https://docs.min.io/minio/baremetal/reference/minio-cli/minio-mc.html) also mimics that.

ripienaar:
We would support `kv rm bucket` and `kv del bucket` as equivalents

miraculli:
I personally don't like it when there are several ways to do the same thing. 
Even when it is not very critical in this case but to get a mental model of the tool someone has to know about this when he writes or reads scripts, docs or tutorials etc.
Even if you just want to understand a script that makes use of `del` and you only learned about `rm` you have to lookup what this `del` is about.

Also... Why are there different commands for `kv` and `object` too? Why not just have a `nats store` command that can be used like `nats store bucket key < foobar.jpg ` to store files/objects?!

Just my thoughts on that...

ripienaar:
kv and obj is singificantly different, would be really confusing to try combine I think.

I hear you about the `rm|del` confusion, it is what it is thogh - we have those commands now but I've had several pieces of feedback saying its confusing.  This proposal means commands will just do the right thing as user intended.

miraculli:
ok.

got it.

miraculli:
I updated this PR to work as suggested. `del` is the primary command now an `rm` is an alias for it.

ripienaar:
Thanks and appologies for the delay.

Do you mind running `go fmt` over your files?

ripienaar:
Meanwhile I combined the kv rm and del commands into one as discussed here

ripienaar:
Thank you

miraculli:
you are welcome...

actually you could also combine `add` / `put` and `info` / `ls` ;-)

PR #322, opened 2022-01-04
Ability to have a delay before sending ack (for --reply and --pull modes) and/or a delay after publishing a message (to generate traffic loads)
Async JS publish error checking and handling

jnmoyne:


cli/bench_command.go
```diff
@@ -455,6 +476,7 @@ func (c *benchCmd) runSubscriber(bm *bench.Benchmark, nc *nats.Conn, startwg *sy
 	mh := func(msg *nats.Msg) {
 		received++
 		if c.reply || (c.js && c.pull) {
+			time.Sleep(c.ackSleep)
```
ripienaar, line 479: need to be careful using this kind of thing in a push consumer - delaying the direct message handler will cause slow consuemrs.  In real world this would be a `SubscribeSync` or this handler would feed into a channel and background go routines would consume and ack to avoid blocking here

jnmoyne: I agree it is best practice to hand off processing to some other go routine through a channel for a push consumer. It would never happen for a push consumer here because the sleep is inside the if statement that is only true for a pull consumer or a replier in request/reply mode.

ripienaar: Ah indeed. All good then LGTM

PR #323, opened 2022-01-07
Fixed usage of code blocks to facilitate copy and paste of the commands

jnmoyne:
Fixed usage of code block showing a shell command being executed and it's output, such that people can simply copy and paste those commands in their own shells without having to then remove the `$ ` prompt at the start of the line. Used 'Output' as a separator, but open to better ideas.


ripienaar:
LGTM

PR #337, opened 2022-01-27
Rename commands for consistency

### LINKED ISSUE #336: Rename commands for consistency
In most places in the CLI where you can list/detail items, the command to list is `ls` and to show information about a specific item `info`:

```
  account info
  consumer info [<flags>] [<stream>] [<consumer>]
  object info <bucket> [<file>]
  server info [<server>]
  stream info [<flags>] [<stream>]
  stream template info [<flags>] [<template>]
```

It seems intuitive that you would then be able to `ls` then use the `info` command for context and schema, but it's `show` rather than `info`:

```
  context show [<flags>] [<name>]
  schema show [<flags>] <schema>
```

For consistency, consider changing all "display information about x" commands to "info". Similarly, list commands are mostly `ls` but there are a few outliers that are `list` instead:

```
  errors list [<flags>] [<match>] [<sort>]
  kv list [<flags>]
  server list [<flags>] [<expect>]
```

It would be [trivial to change](https://github.com/nats-io/natscli/blob/main/cli/context_command.go#L63) but would be a breaking change and might require maintaining compatibility for a few versions with a deprecation notice.

DeanPDX:
Closes #336 

Includes the following changes:

- Use on info vs show
- Use on ls vs list
- Fix minor spelling issue on schema

Commands changed from `show` to `info` were aliased to maintain backward compatibility.

### DIFF
--- a/README.md
+++ b/README.md
@@ -662,7 +662,7 @@ Matched Schemas:
 
 The schemas can be limited using a regular expression, try `nats schema ls request` to see all API requests.
 
-Schemas can be viewed in their raw JSON or YAML formats using `nats schema show io.nats.jetstream.advisory.v1.consumer_action`,
+Schemas can be viewed in their raw JSON or YAML formats using `nats schema info io.nats.jetstream.advisory.v1.consumer_action`,
 these schemas include descriptions about each field and more.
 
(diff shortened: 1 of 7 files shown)


ripienaar:
nice one, thanks a lot for that

PR #347, opened 2022-02-08
powershell installer

scottf:
This is a modified version of install.ps1 found here https://github.com/ConnectEverything/client-tools/pull/2
It's modified to only install the cli, either the stable or nightly. It's been discussed about adding the ability to install a specific release tag.

### DIFF
--- a/install.ps1
+++ b/install.ps1
@@ -35,25 +35,6 @@ Function Read-NightlyVersion() {
 	return $_currentNightly
 }
 
-# Example Url for Read From Url "https://github.com/nats-io/natscli/releases/download/v0.0.28/nats-0.0.28-windows-386.zip"
-Function Read-VersionFromUrl($Url) {
-	$at = $Url.Indexof("download/")
-	$temp = $Url.Substring($at + 9)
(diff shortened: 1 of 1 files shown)


ripienaar:
Seems fine, as a general comment I think its not good that the code doing nightly builds is in a private repo. I can see all you do is run goreleaser but community cant verify that for themselves or inspect how the things they are installing was built /cc @philpennock

philpennock:
There was some back and forth on whether the new installer stuff would also be installing the client tool for the service offered by Synadia, rather than just open source NATS.io tooling, and some uncertainty as to the direction of the new installers as being focused on open source vs service, so I left the repo private initially.

I have no objections to opening it up.

At this time, everything it installs is open source, and the secrets for managing things are properly secured; the README refers to a few service panels which you need authentication to access, with direct links to account IDs, but I think we can live with that.

ripienaar:
OK, I appreciate it's new, happy to park till the picture is clearer - without it holding up this PR, so as long as we aspire to opening that up I think we're good here.

ripienaar:
Should we merge this?

PR #364, opened 2022-03-02
kv upgrade discard policy

scottf:
add checking and updating of discard policy during `kv upgrade <bucket>`


ripienaar (approved):
LGTM, we also have to update the nats.go dependency to ensure buckets made using CLI are right from the start but I will do that tomorrow anyway so dont worry for this PR.

ripienaar:
I upgraded the nats.go dependency in main

PR #365, opened 2022-03-03
Adds durable push consumer and initial KV benchmark mode

jnmoyne:
New flags: --push (for using a push consumer), --consumerbatch (replaces --pullbatch as it's now used for both push and pull consumers), --kv

### DIFF
--- a/cli/bench_command.go
+++ b/cli/bench_command.go
@@ -28,32 +28,36 @@ import (
 )
 
 type benchCmd struct {
-	subject    string
-	numPubs    int
-	numSubs    int
-	numMsg     int
-	msgSize    int
-	csvFile    string
-	noProgress bool
-	request    bool
-	reply      bool
-	syncPub    bool
-	pubBatch   int
-	jsTimeout  time.Duration
-	js         bool
-	storage    string
-	streamName string
-	pull       bool
-	pullBatch  int
-	replicas   int
-	purge      bool
-	ackSleep   time.Duration
-	pubSleep   time.Duration
+	subject       string
+	numPubs       int
+	numSubs       int
+	numMsg        int
+	msgSize       int
+	csvFile       string
+	noProgress    bool
+	request       bool
+	reply         bool
+	syncPub       bool
+	pubBatch      int
+	jsTimeout     time.Duration
+	js            bool
+	storage       string
+	streamName    string
+	pull          bool
+	consumerBatch int
+	replicas      int
+	purge         bool
+	subSleep      time.Duration
+	pubSleep      time.Duration
+	pushDurable   bool
+	consumerName  string
+	kv            bool
+	history       uint8
 }
 
 const (
-	JS_PULLCONSUMER_NAME string = "natscli-benchpull"
-	DEFAULT_STREAM_NAME  string = "benchstream"
+	DEFAULT_DURABLE_CONSUMER_NAME string = "natscli-bench"
+	DEFAULT_STREAM_NAME           string = "benchstream"
 )
 
 func configureBenchCommand(app commandHost) {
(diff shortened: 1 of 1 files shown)


cli/bench_command.go
```diff
@@ -127,17 +136,23 @@ func (c *benchCmd) bench(_ *kingpin.ParseContext) error {
 		return fmt.Errorf("number of messages should be greater than 0")
 	}
 	if c.js && c.numSubs > 0 && c.pull {
-		log.Print("JetStream pull consumer mode: subscriber(s) will explicitly acknowledge the consumption of messages")
+		log.Print("JetStream durable pull consumer mode, subscriber(s) will explicitly acknowledge the consumption of messages")
+	}
+	if c.js && c.numSubs > 0 && c.pushDurable {
+		if c.pull {
+			log.Fatal("the durable consumer must be either pull or push, it can not be both")
```
ripienaar, line 143: should be `The` for consistency, though there's quite a spread of mixing lower/upper case in these log lines, not sure if thats on purpose perhaps?

cli/bench_command.go
```diff
@@ -183,66 +220,96 @@ func (c *benchCmd) bench(_ *kingpin.ParseContext) error {
...
-				return nats.FileStorage
-			case "memory":
-				return nats.MemoryStorage
-			default:
-				{
-					log.Printf("Unknown storage type %s, using memory", c.storage)
-					return nats.MemoryStorage
+		if c.kv {
+			// create bucket
+			_, err := js.CreateKeyValue(&nats.KeyValueConfig{Bucket: c.subject, History: c.history, Storage: nats.MemoryStorage, Description: "nats bench bucket", Replicas: c.replicas})
```
ripienaar, line 227: should also use `storageType` here?

cli/bench_command.go
```diff
@@ -183,66 +220,96 @@ func (c *benchCmd) bench(_ *kingpin.ParseContext) error {
...
+						ReplayPolicy:   nats.ReplayInstantPolicy,
+						MaxAckPending:  c.consumerBatch * c.numSubs,
+					})
+					if err != nil {
+						log.Fatal("error creating the durable push consumer: ", err)
+					}
+					defer func() {
+						err := js.DeleteConsumer(c.streamName, c.consumerName)
+						if err != nil {
+							log.Printf("error deleting the durable push consumer on stream %s: %v", c.streamName, err)
```
ripienaar, line 294: should be fatal?

ripienaar (changes requested):
few small comments, in general I think these methods are absolute monsters now, next time we add some testing modes we should consider splitting them into smaller functions perhaps

cli/bench_command.go
jnmoyne: Sure, changed it (though it doesn't make much difference since this is cleanup code)

cli/bench_command.go
jnmoyne: I homogenized so now all error messages start with a capital letter. Some people told me 'in Go error messages should not start with an upper case' but I since checked and that's only for Errorf to cover the case the error message in question is being bubbled up the stack where it is then included in the error log (i.e. `log.Fatalf("This error happened: %v\n",err)` so the error string should not start with a capital). In this case here my error messages are all log.Fatalf directly so they should all be starting with upper case. (source: https://github.com/golang/go/wiki/CodeReviewComments#error-strings)

cli/bench_command.go
jnmoyne: You are correct, thanks for the catch!

cli/bench_command.go
ripienaar: correct, where they are essentially the final result of the error as logged is shown to the user it is fine to be capital, but if say you logged the error value `log.Fatalf("Thing failed: %v", err)` its better error strings are lower case.

Anyway, i dont care if its upper or lower I care more about consistency, consistently "incorrect" throughout is actually better than being a mix of things :)

ripienaar (approved):
LGTM

PR #384, opened 2022-03-18
add a completion format output for ctx list

ripienaar:


ripienaar:
```
$ nats context ls --completion
choria:Choria
choria.system:Choria System
leaf:leaf domain
lon*:London
nyc:New York
sfo:San Francisco
system:System User - Cluster
system.lon:System User - London
system.nyc:System User - New York
system.sfo:System User - San Francisco
test\:test\:test:London
weather.london:London Weather
```

ripienaar:
I suppose I should also escape `\` in both strings right?

PR #393, opened 2022-04-01
Add max-pending flag for 'consumer edit' command

boris-ilijic:
Using `consumer add` command it is possible to specify `--max-pending` flag.
This PR enables the same but for `consumer edit` flow. Currently, `consumer edit` command doesn't allow to change `max-pending` property of JS Consumer.


ripienaar:
good catch, thank you. I shold do this the same as for stream edit so the options are not duplicated

PR #404, opened 2022-04-05
Add publish request support for multiple responses

### LINKED ISSUE #403: Add an option to make a request and receive multiple replies
Currently both `nats req` and `nats pub --wait` exit after one reply received.  This prevents usage/demonstration of voting/multi-ping pattern of service request (multiple service responders receive the request and individually reply).

Propose adding a _reply wait duration_ option such that any inbox replies received up to the expiry of reply wait duration are printed by CLI.

ColinSullivan1:
Adds a publish request wait to allow for multiple responses from a publish call.  Updated to honor the timeout for the first message, then wait a `reply-timeout` for subsequent messages, a suggestion by @derekcollison for better UX.

`--replies 0` will gather responses until the global timeout is reached.

### Example

```text
nats reply foo "response from service a" --queue a &
nats reply foo "response from service b" --queue b &
nats reply foo "response from service c" --queue c &
```

Publish and wait (edited)

```
$ nats req foo hello
11:15:03 Sending request on "foo"
11:15:03 Received on "_INBOX.Wu6eONq021mLlGG5nWuMGp.SqEPt48b" rtt 406.868µs
response from service a

$ nats req --replies 2 foo hello
11:15:03 Sending request on "foo"
11:15:03 Received on "_INBOX.qf0sz6O8QzaQR3hp2NEfIC.AQNxkIEZ" rtt 277.961µs
response from service b

11:15:03 Received on "_INBOX.qf0sz6O8QzaQR3hp2NEfIC.AQNxkIEZ" rtt 292.951µs
response from service c

$ nats req --count 2 --replies 2 foo hello
11:15:03 Sending request on "foo"
11:15:03 Received on "_INBOX.CxwBcyiUTWqGKVDNTRhivV.zgA2BcQD" rtt 286.797µs
response from service b

11:15:03 Received on "_INBOX.CxwBcyiUTWqGKVDNTRhivV.zgA2BcQD" rtt 332.823µs
response from service c

11:15:03 Sending request on "foo"
11:15:03 Received on "_INBOX.CxwBcyiUTWqGKVDNTRhivV.QCt1hPjN" rtt 204.034µs
response from service c

11:15:03 Received on "_INBOX.CxwBcyiUTWqGKVDNTRhivV.QCt1hPjN" rtt 212.46µs
response from service b

$ nats req --replies 4 foo hello
11:15:03 Sending request on "foo"
11:15:03 Received on "_INBOX.xbJ7N0Fn7DjrqGhUncDuJi.03IZvLTk" rtt 283.41µs
response from service c

11:15:03 Received on "_INBOX.xbJ7N0Fn7DjrqGhUncDuJi.03IZvLTk" rtt 296.587µs
response from service b

11:15:03 Received on "_INBOX.xbJ7N0Fn7DjrqGhUncDuJi.03IZvLTk" rtt 300.663µs
response from service a

$ nats req --replies 0 --timeout 2s foo hello
11:15:03 Sending request on "foo"
11:15:03 Received on "_INBOX.ed9PxZ7jHPwOcAV6vivU2s.0zCVrCFU" rtt 334.656µs
response from service c

11:15:03 Received on "_INBOX.ed9PxZ7jHPwOcAV6vivU2s.0zCVrCFU" rtt 350.819µs
response from service a

11:15:03 Received on "_INBOX.ed9PxZ7jHPwOcAV6vivU2s.0zCVrCFU" rtt 355.099µs
response from service b

$ nats req --replies 4 --reply-timeout 1s foo hello
11:15:05 Sending request on "foo"
11:15:05 Received on "_INBOX.eUkVIs3ld62Uu1LzLtrrCR.CFE2ojkB" rtt 286µs
response from service c

11:15:05 Received on "_INBOX.eUkVIs3ld62Uu1LzLtrrCR.CFE2ojkB" rtt 303.916µs
response from service b

11:15:05 Received on "_INBOX.eUkVIs3ld62Uu1LzLtrrCR.CFE2ojkB" rtt 310.99µs
response from service a

$ nats req --replies 4 --reply-timeout 1us foo hello
11:15:06 Sending request on "foo"
11:15:06 Received on "_INBOX.7uSdKcRuPFxBsIEq3KTVdb.uCUmA2bZ" rtt 286.172µs
response from service a

11:15:06 Received on "_INBOX.7uSdKcRuPFxBsIEq3KTVdb.uCUmA2bZ" rtt 300.743µs
response from service b

11:15:06 Received on "_INBOX.7uSdKcRuPFxBsIEq3KTVdb.uCUmA2bZ" rtt 304.645µs
response from service c

$ nats req --count 2 --replies 0 --timeout 2s foo hello
11:15:06 Sending request on "foo"
11:15:06 Received on "_INBOX.lASTUVgiH8vp0WjGItD2FZ.vXN8TArl" rtt 298.671µs
response from service a

11:15:06 Received on "_INBOX.lASTUVgiH8vp0WjGItD2FZ.vXN8TArl" rtt 315.217µs
response from service b

11:15:06 Received on "_INBOX.lASTUVgiH8vp0WjGItD2FZ.vXN8TArl" rtt 319.691µs
response from service c

11:15:08 Sending request on "foo"
11:15:08 Received on "_INBOX.lASTUVgiH8vp0WjGItD2FZ.79AY5Nin" rtt 356.559µs
response from service c

11:15:08 Received on "_INBOX.lASTUVgiH8vp0WjGItD2FZ.79AY5Nin" rtt 369.079µs
response from service b

11:15:08 Received on "_INBOX.lASTUVgiH8vp0WjGItD2FZ.79AY5Nin" rtt 373.056µs
response from service a

$ nats req --timeout 2s invalid hello
11:15:10 Sending request on "invalid"
11:15:10 No responders are available

$ nats req --timeout 1us --count 4 foo hello
11:15:10 Sending request on "foo"
11:15:10 Sending request on "foo"
11:15:10 Sending request on "foo"
11:15:10 Sending request on "foo"
```

Resolves #403 

Signed-off-by: Colin Sullivan <colin@synadia.com>

### DIFF
--- a/cli/pub_command.go
+++ b/cli/pub_command.go
@@ -118,7 +118,7 @@ Available template functions are:
 	req.Flag("header", "Adds headers to the message").Short('H').StringsVar(&c.hdrs)
 	req.Flag("count", "Publish multiple messages").Default("1").IntVar(&c.cnt)
 	req.Flag("replies", "Wait for multiple replies from services. 0 waits until timeout").Default("1").IntVar(&c.replyCount)
-	req.Flag("reply-timeout", "Maximum timeout between incoming replies.").Default("250ms").DurationVar(&c.replyTimeout)
+	req.Flag("reply-timeout", "Maximum timeout between incoming replies.").Default("300ms").DurationVar(&c.replyTimeout)
 
 }
 
@@ -141,7 +141,7 @@ func (c *pubCmd) doReq(nc *nats.Conn, progress *uiprogress.Bar) error {
 			log.Printf("Sending request on %q\n", c.subject)
 		}
 
-		body, err := pubReplyBodyTemplate(c.body, 0)
+		body, err := pubReplyBodyTemplate(c.body, i)
 		if err != nil {
 			log.Printf("Could not parse body template: %s", err)
 		}
@@ -173,22 +173,23 @@ func (c *pubCmd) doReq(nc *nats.Conn, progress *uiprogress.Bar) error {
 		// loop until reply count is met, or if zero, until we
 		// timeout receiving messages.
 		rc := 0
+		var rttAg time.Duration
 		for {
 			m, err := s.NextMsg(timeout)
 			if err != nil {
 				if err == nats.ErrTimeout {
-					return nil
+					// continue to publish additional messages.
+					break
 				}
 				if err == nats.ErrNoResponders {
 					log.Printf("No responders are available")
 					return nil
 				}
-
-				log.Printf("Error receiving message: %v", err)
 				return err
 			}
 
-			log.Printf("Received on %q rtt %v", m.Subject, time.Since(start))
+			rtt := time.Since(start)
+			log.Printf("Received on %q rtt %v", m.Subject, rtt)
 
 			if len(m.Header) > 0 {
 				for h, vals := range m.Header {
@@ -208,7 +209,16 @@ func (c *pubCmd) doReq(nc *nats.Conn, progress *uiprogress.Bar) error {
 			if c.replyCount > 0 && rc == c.replyCount {
 				break
 			}
-			timeout = c.replyTimeout
+
+			if c.replyCount == 0 {
+				// if we are waiting for the general timeout then
+				// calculate remaining
+				timeout = opts.Timeout - time.Since(start)
+			} else {
+				// Otherwise, use the average response deltas
+				rttAg += rtt
+				timeout = rttAg/time.Duration(rc) + c.replyTimeout
+			}
 		}
 
 		// Unsubscribe for the unbound case, NOOP is already auto unsubscribed.
--- a/cli/pub_command.go
+++ b/cli/pub_command.go
@@ -36,6 +36,7 @@ type pubCmd struct {
 	hdrs       []string
 	cnt        int
 	sleep      time.Duration
+	waitDur    time.Duration
 	forceStdin bool
 }
 
@@ -67,6 +68,7 @@ Available template functions are:
 	pub.Arg("subject", "Subject to subscribe to").Required().StringVar(&c.subject)
 	pub.Arg("body", "Message body").Default("!nil!").StringVar(&c.body)
 	pub.Flag("wait", "Wait for a reply from a service").Short('w').BoolVar(&c.req)
+	pub.Flag("wait-duration", "Wait for multiple replies from a service").Short('d').Default("2s").DurationVar(&c.waitDur)
 	pub.Flag("reply", "Sets a custom reply to subject").StringVar(&c.replyTo)
 	pub.Flag("header", "Adds headers to the message").Short('H').StringsVar(&c.hdrs)
 	pub.Flag("count", "Publish multiple messages").Default("1").IntVar(&c.cnt)
@@ -127,6 +129,73 @@ func (c *pubCmd) prepareMsg(body []byte, seq int) (*nats.Msg, error) {
 	msg.Data = body
 
 	return msg, parseStringsToMsgHeader(c.hdrs, seq, msg)
+}
+
+func (c *pubCmd) doReqWait(nc *nats.Conn, progress *uiprogress.Bar) error {
+	for i := 1; i <= c.cnt; i++ {
+		start := time.Now()
+
+		if !c.raw && progress == nil {
+			log.Printf("Sending request on %q\n", c.subject)
+		}
+
+		body, err := pubReplyBodyTemplate(c.body, 0)
+		if err != nil {
+			log.Printf("Could not parse body template: %s", err)
+		}
+
+		msg, err := c.prepareMsg(body, 0)
+		if err != nil {
+			return err
+		}
+		msg.Reply = nc.NewRespInbox()
+
+		s, err := nc.Subscribe(msg.Reply, func(m *nats.Msg) {
+			log.Printf("Received on %q rtt %v", m.Subject, time.Since(start))
+
+			if len(m.Header) > 0 {
+				for h, vals := range m.Header {
+					for _, val := range vals {
+						log.Printf("%s: %s", h, val)
+					}
+				}
+
+				fmt.Println()
+			}
+
+			fmt.Println(string(m.Data))
+			if !strings.HasSuffix(string(m.Data), "\n") {
+				fmt.Println()
+			}
+		})
+		if err != nil {
+			return err
+		}
+		defer s.Unsubscribe()
+
+		err = nc.PublishMsg(msg)
+		if err != nil {
+			return err
+		}
+
+		// cls - TODO maybe add response count and select with
+		// channel as well to short circuit if response count
+		// is known apriori.
+		time.Sleep(c.waitDur)
+
+		// we don't want any responses after we wait.
+		s.Unsubscribe()
+
+		// include a publish sleep in the wait duration
+		if c.cnt > 1 && c.sleep > 0 {
+			st := c.sleep - c.waitDur
+			if st > 0 {
+				time.Sleep(st)
+			}
+		}
+	}
+
+	return nil
 }
 
 func (c *pubCmd) doReq(nc *nats.Conn, progress *uiprogress.Bar) error {
@@ -212,6 +281,10 @@ func (c *pubCmd) publish(_ *kingpin.ParseContext) error {
 		fmt.Println()
 		uiprogress.Start()
 		defer func() { uiprogress.Stop(); fmt.Println() }()
+	}
+
+	if c.waitDur != 0 {
+		return c.doReqWait(nc, progress)
 	}
 
 	if c.req {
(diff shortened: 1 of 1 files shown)


tbeets:
Would be cool to add max receives to max duration (whichever happens first).  

Here's an example riffing on your PR: https://github.com/tbeets/gonats-101/blob/main/nats-req-multi/main.go

ColinSullivan1:
Yep, cool - or thinking about it, could expand this with a count and/or handle timeout... e.g. https://gist.github.com/r4um/c1ab51b8757fc2d75d30320933cdbdf6

cli/pub_command.go
```diff
@@ -129,6 +131,73 @@ func (c *pubCmd) prepareMsg(body []byte, seq int) (*nats.Msg, error) {
 	return msg, parseStringsToMsgHeader(c.hdrs, seq, msg)
 }
 
+func (c *pubCmd) doReqWait(nc *nats.Conn, progress *uiprogress.Bar) error {
```
ripienaar, line 134: in order to decrease the duplication I think we can always use the same req - this one - we just need to know to exit after first message when wait duration is 0

The existing `doReqAsync()` function has a lot of this logic in it, including for example early short circuit when no responses are recieved for 300ms etc. 

So after @tbeets opened the issue my thought was that I would make that a bit more generic - or extract from it it's core 1:n reply handling to use here.  Thats a bit of a chore, happy to do that on Friday.

ColinSullivan1: I'll take a look, thanks @ripienaar.

ColinSullivan1:
After some thought I simplified the request a bit, de-duped code based on feedback, and used a select like @tbeets suggested.  Also removed one parameter so we're just left with `--wait-multiple`, and the timeout parameter is used instead of wait-duration.  Zero will wait until the timeout.  No responders are also covered.

I looked into reusing the code in util, that would be major work as you mentioned @ripienaar.  This seemed simpler and easier given support of the pub count and sleep options in the meantime.

cli/pub_command.go
```diff
@@ -66,7 +67,8 @@ Available template functions are:
 	pub := app.Command("publish", pubHelp).Alias("pub").Action(c.publish)
 	pub.Arg("subject", "Subject to subscribe to").Required().StringVar(&c.subject)
 	pub.Arg("body", "Message body").Default("!nil!").StringVar(&c.body)
-	pub.Flag("wait", "Wait for a reply from a service").Short('w').BoolVar(&c.req)
+	pub.Flag("wait", "Wait for a single reply from a service").Short('w').BoolVar(&c.req)
+	pub.Flag("wait-multiple", "Wait for multiple replies from services. 0 waits until timeout").Default("1").IntVar(&c.waitCount)
```
ripienaar, line 71: Wouldnt it be more logical if `nats req` just had an option to wait for multiple responses or a set period keeping `nats pub` as sending only? Having `nats pub` gain some `req` features that `req` doesnt have seems weird to me. 

While these 2 just reuse the same logic and is so close together the line between publish only and interact with services is pretty strong so would want to keep that

Rest of the PR is great thank you

ripienaar: Yes I have the delta code in tbe other function I mentioned. If we need to go all in and do that might still be worth making that a bit more generic for reuse

tbeets: @ColinSullivan1 The `--wait-multiple` option on `nats pub` was elegant.  What are you thinking for `nats req` -- something like just `--multiple`?

ripienaar: Don’t think delta based early termination makes sense here. It makes sense for server api interactions like varz because we know servers respond immediately. We do not know that for some random service.

ColinSullivan1: I went ahead and added some very simple delta-like code upon request, and that behavior can be overridden with `--request 0`.  Certainly open for discussion or suggestions.  Also moved to the sync sub - it was simpler.

ColinSullivan1: I updated the description to reflect the new usage, and the timestamps give an indication of UX.

derekcollison:
I will let @ripienaar make the call on the code but looks pretty good to me, well done.

cli/pub_command.go
```diff
@@ -66,7 +68,7 @@ Available template functions are:
 	pub := app.Command("publish", pubHelp).Alias("pub").Action(c.publish)
 	pub.Arg("subject", "Subject to subscribe to").Required().StringVar(&c.subject)
 	pub.Arg("body", "Message body").Default("!nil!").StringVar(&c.body)
-	pub.Flag("wait", "Wait for a reply from a service").Short('w').BoolVar(&c.req)
+	pub.Flag("wait", "Wait for a single reply from a service").Short('w').BoolVar(&c.req)
```
ripienaar, line 71: we should probably just remove this option entirely, think its here as a old leftover or just a mistake. pub should just pub.

cli/pub_command.go
```diff
@@ -115,6 +117,9 @@ Available template functions are:
 	req.Flag("raw", "Show just the output received").Short('r').Default("false").BoolVar(&c.raw)
 	req.Flag("header", "Adds headers to the message").Short('H').StringsVar(&c.hdrs)
 	req.Flag("count", "Publish multiple messages").Default("1").IntVar(&c.cnt)
+	req.Flag("replies", "Wait for multiple replies from services. 0 waits until timeout").Default("1").IntVar(&c.replyCount)
+	req.Flag("reply-timeout", "Maximum timeout between incoming replies.").Default("250ms").DurationVar(&c.replyTimeout)
```
ripienaar, line 121: If we are making this default we need to somehow explain to users that we are using this and not --timeout in the UX of the application.

As I mentioned earlier I think for interacting with nats server APIs this would be a good default - they respond immediately - but we have no idea what services users interact with and this default could be confusing them not knowing why the default timeout is being ignored for example. Even the example in the docs again wttr using nats reply would not be reliable since that API is far latency wise and might not respond immediately.  

So not sure, we should maybe warn at the end when early terminating based on prefering that value over --timeout

cli/pub_command.go
```diff
@@ -130,13 +135,13 @@ func (c *pubCmd) prepareMsg(body []byte, seq int) (*nats.Msg, error) {
...
 func (c *pubCmd) doReq(nc *nats.Conn, progress *uiprogress.Bar) error {
+
 	for i := 1; i <= c.cnt; i++ {
-		start := time.Now()
 		if !c.raw && progress == nil {
 			log.Printf("Sending request on %q\n", c.subject)
 		}
 
-		body, err := pubReplyBodyTemplate(c.body, i)
+		body, err := pubReplyBodyTemplate(c.body, 0)
```
ripienaar, line 144: I think the `0` here should be `i`, what was the reason for the change? 

This is the value that `{{Count}}` would set in the body template and its the number of messages sent, setting this to 0 always will break that feature.

cli/pub_command.go
```diff
@@ -146,16 +151,43 @@ func (c *pubCmd) doReq(nc *nats.Conn, progress *uiprogress.Bar) error {
...
+				if err == nats.ErrTimeout {
+					return nil
+				}
+				if err == nats.ErrNoResponders {
+					log.Printf("No responders are available")
+					return nil
+				}
+
+				log.Printf("Error receiving message: %v", err)
+				return err
```
ripienaar, line 188: this will log here and in the exit handler wont it? Error would be logged twice? maybe better `return fmt.Error("could not receive message: %v", err)` which would then be handled and logged higher up

ripienaar (changes requested):
some minor tweeks, looking good!

cli/pub_command.go
ColinSullivan1: That was a bug from an earlier iteration - thanks for catching that - will fix.

cli/pub_command.go
ColinSullivan1: Per our slack convo, I'll update the subsequent timeout to do 300ms + response time for the first.  So, for example if the first response came in at 100ms, then we'd wait 400ms (100ms + 300ms) for the 2nd, resulting a total of 500ms for the entire operation.

ColinSullivan1: Update, will use a rtt average.

ColinSullivan1:
I've added some dynamic averages around rtt for slower systems, so we can have a good UX for systems that respond quickly, and work with systems that have larger RTTs.  If there is a single service that is very slow to respond then that response may be lost.  However, that is also what `--replies 0` with a long timeout would handle.

ripienaar (changes requested):

cli/pub_command.go
```diff
@@ -146,39 +151,87 @@ func (c *pubCmd) doReq(nc *nats.Conn, progress *uiprogress.Bar) error {
...
+				}
+				if err == nats.ErrNoResponders {
+					log.Printf("No responders are available")
+					return nil
+				}
+				return err
+			}
+
+			rtt := time.Since(start)
+			log.Printf("Received on %q rtt %v", m.Subject, rtt)
```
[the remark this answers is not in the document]
ColinSullivan1: It was in the original request API.  I can go either way - wdyt?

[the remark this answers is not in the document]
ripienaar: ok we leave as is for now :)

ripienaar (approved):
LGTM

PR #429, opened 2022-04-16
Humanize stream byte parameters

ColinSullivan1:
Set a default stream size of 256MB and humanize the stream parameters (stream size, max msgs size).

```text
$ nats s create colin
? Subjects colin
? Storage file
? Replication 1
? Retention Policy Limits
? Discard Policy Old
? Stream Messages Limit -1
? Per Subject Messages Limit -1
? Total Stream Size 256MB
? Message TTL -1
? Max Message Size -1
? Duplicate tracking time window 2m0s
? Allow message Roll-ups No
? Allow message deletion Yes
? Allow purging subjects or the entire stream Yes
Stream colin was created

Information for Stream colin created 2022-04-16T17:52:17-06:00

Configuration:

             Subjects: colin
     Acknowledgements: true
            Retention: File - Limits
             Replicas: 1
       Discard Policy: Old
     Duplicate Window: 2m0s
    Allows Msg Delete: true
         Allows Purge: true
       Allows Rollups: false
     Maximum Messages: unlimited
        Maximum Bytes: 244 MiB
          Maximum Age: unlimited
 Maximum Message Size: unlimited
    Maximum Consumers: unlimited


Cluster Information:

                 Name: ngsprod-aws-uswest2
               Leader: aws-uswest2-natsjs-1

State:

             Messages: 0
                Bytes: 0 B
             FirstSeq: 0
              LastSeq: 0
     Active Consumers: 0
```

Editing:

```
$ nats s edit colin --max-bytes=10MB
Differences (-old +new):
  api.StreamConfig{
  	... // 5 identical fields
  	MaxMsgsPer: -1,
  	MaxMsgs:    -1,
- 	MaxBytes:   256000000,
+ 	MaxBytes:   10000000,
  	MaxAge:     s"0s",
  	MaxMsgSize: -1,
  	... // 13 identical fields
  }
? Really edit Stream colin Yes
Stream colin was updated

Information for Stream colin created 2022-04-16T17:52:17-06:00

Configuration:

             Subjects: colin
     Acknowledgements: true
            Retention: File - Limits
             Replicas: 1
       Discard Policy: Old
     Duplicate Window: 2m0s
    Allows Msg Delete: true
         Allows Purge: true
       Allows Rollups: false
     Maximum Messages: unlimited
        Maximum Bytes: 9.5 MiB
          Maximum Age: unlimited
 Maximum Message Size: unlimited
    Maximum Consumers: unlimited


Cluster Information:

                 Name: ngsprod-aws-uswest2
[...]

PR #432, opened 2022-04-17
Humanize byte arguments on the cli and prompts

ripienaar:
This uses a bytes parser derived from, and compatible with,
the server options parser but additionally supports TB, tb and
t as aliases for T. All are 1024 based.

We use that in cli arguments and update all other bytes
parsing to be the same, this means 1GB is consistent with
the rendering of 1GiB - all 1024 based.

Do the same on stream, kv and object.

Fixes a bug that required a size be set on accounts
that do not require it and default to 256MB when it is
required

### DIFF
--- a/cli/stream_command.go
+++ b/cli/stream_command.go
@@ -60,8 +60,10 @@ type streamCmd struct {
 	storage               string
 	maxMsgLimit           int64
 	maxMsgPerSubjectLimit int64
+	maxBytesLimitString   string
 	maxBytesLimit         int64
 	maxAgeLimit           string
+	maxMsgSizeString      string
 	maxMsgSize            int64
 	maxConsumers          int
 	reportSortConsumers   bool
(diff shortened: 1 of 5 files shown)


ripienaar:
This ended up a bit more involved than what @ColinSullivan1 did in https://github.com/nats-io/natscli/pull/429 but I think the outcome is better, limits remain ints so keeping the general fallout from the change smaller and parsing of cli/user supplied is improved and handle many more cases than we did before. all 1024 based avoiding the weirdness of why does `1M` end up being less than `1M`.

cli/stream_command.go
```diff
@@ -141,9 +143,9 @@ func configureStreamCommand(app commandHost) {
 		f.Flag("dupe-window", "Window size for duplicate tracking").Default("").StringVar(&c.dupeWindow)
 		f.Flag("json", "Produce JSON output").Short('j').BoolVar(&c.json)
 		f.Flag("max-age", "Maximum age of messages to keep").Default("").StringVar(&c.maxAgeLimit)
-		f.Flag("max-bytes", "Maximum bytes to keep").Int64Var(&c.maxBytesLimit)
+		f.Flag("max-bytes", "Maximum bytes to keep").StringVar(&c.maxBytesLimitString)
```
jnmoyne, line 146: Why not .Default("256MB") there?

ripienaar: Only 256 default when the account requires a size. Not always.

derekcollison:
Possible to make it such that when I say `2g` as max bytes that when I subsequently do a `stream info` I see 2g? Can we force 2g -> 2gib?

ripienaar:
> Possible to make it such that when I say `2g` as max bytes that when I subsequently do a `stream info` I see 2g? Can we force 2g -> 2gib?

I am sure it’s in theory possible but would be awkward. I also quite like that you get a standardised confirmation that your input was correctly parsed

derekcollison:
I will just learn to type 2gib ;)

ripienaar:
G, gib, GiB, GB all means the same now when entering - wrong but least surprising so you can do whichever you like :)

ColinSullivan1 (approved):
LGTM.  Thanks @ripienaar!

PR #438, opened 2022-04-22
add maxbyte arg to bench and small fixes/improvements to account info and stream and consumer commands

jnmoyne:


cli/account_command.go
```diff
@@ -236,12 +237,14 @@ func (c *actCmd) infoAction(_ *kingpin.ParseContext) error {
 		fmt.Printf("Account Usage:\n\n")
 		fmt.Printf("    Storage: %s\n", humanize.IBytes(info.Store))
 		fmt.Printf("     Memory: %s\n", humanize.IBytes(info.Memory))
-		fmt.Printf("    Streams: %s\n", humanize.IBytes(uint64(info.Streams)))
-		fmt.Printf("  Consumers: %s\n", humanize.IBytes(uint64(info.Consumers)))
+		fmt.Printf("    Streams: %d\n", info.Streams)
```
ripienaar, line 240: But maybe humanize Comma now?

cli/account_command.go
```diff
@@ -236,12 +237,14 @@ func (c *actCmd) infoAction(_ *kingpin.ParseContext) error {
...
+		fmt.Printf("  Consumers: %d\n", info.Consumers)
 		fmt.Println()
 
 		fmt.Printf("Account Limits:\n\n")
 
+		fmt.Printf("   Max Message Payload: %s (%d bytes) \n\n", humanize.IBytes(uint64(nc.MaxPayload())), nc.MaxPayload())
```
ripienaar, line 246: I think drop the printing of exact bytes. We never do that anywhere else I think?

cli/bench_command.go
```diff
@@ -63,31 +66,32 @@ const (
 func configureBenchCommand(app commandHost) {
 	c := &benchCmd{}
 	bench := app.Command("bench", "Benchmark utility").Action(c.bench)
-	bench.Arg("subject", "Subject to use for testing").Required().StringVar(&c.subject)
+	bench.Arg("subject", "Subject to use for the benchmark").Required().StringVar(&c.subject)
 	bench.Flag("pub", "Number of concurrent publishers").Default("0").IntVar(&c.numPubs)
 	bench.Flag("sub", "Number of concurrent subscribers").Default("0").IntVar(&c.numSubs)
+	bench.Flag("js", "Use JetStream streaming").Default("false").BoolVar(&c.js)
```
ripienaar, line 72: “JetStream Streaming” is that really a way we are referring to it?  I think in docs it’s mostly just JetStream so reads a bit odd here

cli/bench_command.go
```diff
@@ -63,31 +66,32 @@ const (
...
+	bench.Flag("request", "Request/Reply mode: publishers send requests waits for a reply").Default("false").BoolVar(&c.request)
+	bench.Flag("reply", "Request/Reply mode: subscribers send replies").Default("false").BoolVar(&c.reply)
```
ripienaar, line 74: I also prefer Request/Repmy but I think nats docs etc mostly say Request-Reply

cli/bench_command.go
```diff
@@ -63,31 +66,32 @@ const (
...
 	bench.Flag("msgs", "Number of messages to publish").Default("100000").IntVar(&c.numMsg)
-	bench.Flag("size", "Size of the test messages").Default("128").IntVar(&c.msgSize)
-	bench.Flag("csv", "Save benchmark data to CSV file").StringVar(&c.csvFile)
+	bench.Flag("size", "Size of the test messages").Default("128").StringVar(&c.msgSizeString)
 	bench.Flag("no-progress", "Disable progress bar while publishing").Default("false").BoolVar(&c.noProgress)
+	bench.Flag("csv", "Save benchmark data to CSV file").StringVar(&c.csvFile)
+	bench.Flag("purge", "Purge the stream before running").Default("false").BoolVar(&c.purge)
+	bench.Flag("storage", "JetStream storage (memory/file) for the \"benchstream\" stream").Default("memory").EnumVar(&c.storage, "memory", "file")
+	bench.Flag("replicas", "Number of stream replicas for the \"benchstream\" stream").Default("1").IntVar(&c.replicas)
+	bench.Flag("maxbytes", "The maximum size of the stream or KV bucket in bytes (1GB by default)").Default("1GB").StringVar(&c.streamMaxBytesString)
```
ripienaar, line 83: —help already shows the default

cli/bench_command.go
```diff
@@ -215,25 +233,33 @@ func (c *benchCmd) bench(_ *kingpin.ParseContext) error {
...
+			log.Fatalf("NATS connection failed %s", err)
 		}
 
 		js, err = nc.JetStream(nats.MaxWait(c.jsTimeout))
 		if err != nil {
-			log.Fatalf("Couldn't get the JetStream context: %v", err)
+			log.Fatalf("Couldn't get the JetStream context %v", err)
 		}
 		if c.kv {
+			// There is no way to purge all the keys in a KV bucket in a single operation so deleting the bucket instead
```
ripienaar, line 244: Normal stream purge?

cli/bench_command.go
```diff
@@ -489,12 +516,12 @@ func jsPublisher(c benchCmd, nc *nats.Conn, progress *uiprogress.Bar, msg []byte
...
 	js, err := nc.JetStream()
 	if err != nil {
-		log.Fatalf("Couldn't get the JetStream context: %v", err)
+		log.Fatalf("Couldn't get the JetStream context %v", err)
 	}
 
 	kvBucket, err := js.KeyValue(c.subject)
 	if err != nil {
-		log.Fatalf("Couldn't find kv store %s: %v", c.subject, err)
+		log.Fatalf("Couldn't find kv store %s %v", c.subject, err)
```
ripienaar, line 524: Generally people would write go errors in a way that they would make sense after “:”. Some might be weird without. Especially some that are a chain or “:” sep errors. 

Agree though here where it’s user shown the : thing is annoying. But I found it weird both ways :)

cli/stream_command.go
```diff
@@ -136,61 +136,55 @@ func configureStreamCommand(app commandHost) {
...
 		f.Flag("discard", "Defines the discard policy (new, old)").EnumVar(&c.discardPolicy, "new", "old")
-		f.Flag("dupe-window", "Window size for duplicate tracking").Default("").StringVar(&c.dupeWindow)
-		f.Flag("json", "Produce JSON output").Short('j').BoolVar(&c.json)
 		f.Flag("max-age", "Maximum age of messages to keep").Default("").StringVar(&c.maxAgeLimit)
 		f.Flag("max-bytes", "Maximum bytes to keep").StringVar(&c.maxBytesLimitString)
 		f.Flag("max-consumers", "Maximum number of consumers to allow").Default("-1").IntVar(&c.maxConsumers)
 		f.Flag("max-msg-size", "Maximum size any 1 message may be").StringVar(&c.maxMsgSizeString)
 		f.Flag("max-msgs", "Maximum amount of messages to keep").Default("0").Int64Var(&c.maxMsgLimit)
 		f.Flag("max-msgs-per-subject", "Maximum amount of messages to keep per subject").Default("0").Int64Var(&c.maxMsgPerSubjectLimit)
+		f.Flag("dupe-window", "Window size for duplicate tracking").Default("").StringVar(&c.dupeWindow)
```
ripienaar, line 157: Should we clarify it’s seconds or a duration?

ripienaar (changes requested):
Many comments sorry :)

cli/bench_command.go
jnmoyne: I meant no way through the Go kv API. There is a kv.Purge but it it's different (works on a key not the whole bucket). You're right it's more efficient to purge the underlying stream, but since that assumes you know the convention to make the stream name from the bucket name, still think it would be better for nats.go to have a 'kv.PurgeAll()' (or PurgeBucket() ?) call that does the purge of the underlying stream while isolating the user to have to know that naming convention. Should I create an issue for that on nats.go?

cli/bench_command.go
jnmoyne: Ah yes forgot that, thanks for reminder

cli/bench_command.go
jnmoyne: with a / is how I have always written it (for many decades now :)), but you are right that the docs have a mixture of - and /. As this made me curious and I did some research about it and as it turns out the hyphenation is actually correct here because because it represents the interaction of a request _and_ a reply while the slash is the one to use to convey that it can be a request _or_ a reply.
So you learn something new every day and I will go with the hyphen (and also harmonize the docs while I'm at it)

cli/bench_command.go
jnmoyne: I think it's an attempt at distinguishing --js (streaming benchmark) vs --kv (kv benchmarking) buy yeah it ends up reading a bit odd so cleaning it up now.

cli/bench_command.go
jnmoyne: I changed this because the errors coming back from nats.go start with 'nats: xxxxx' so if you want to add some context ahead of the nats.go error (like here repeating the subject) and end that with a ':' you end up with a chain indeed (.e.g. 'nats: error: nats: bucket not found') which looked weird  and annoying to me so I took mine out. Didn't realize it was the convention to have this chain of : to delineate the layers of the onion and therefore you must always end on a :  so reverting to doing that.

cli/stream_command.go
jnmoyne: Changed to: "Duration of the duplicate message tracking window"

cli/account_command.go
jnmoyne: Sure, I had added that while testing something and forgot about it.

cli/account_command.go
jnmoyne: Great suggestion

cli/bench_command.go
ripienaar: Yeah I think that was in the initial spec - however you can see in kv info how to get the real jetstream stream info with name and all for a traditional purge.

Doesnt matter its not like this is wrong, just a note, so dont need to change

ripienaar:
Looks good. If you pin static check to their latest release it should pass tests. 

I had to take master to support go 1.18 and not master doesn’t do 1.17 anymore hehe.

cli/account_command.go
jnmoyne: Actually I remember now, I added the exact bytes because that is what `nsc describe account` does.

cli/bench_command.go
jnmoyne: Right I changed it to do the purge by composing the underlying stream name directly from the subject name, and will file an issue in nats.go because I think that call is missing.

PR #442, opened 2022-04-28
[add] account restore to match account backup

matthiashanel:
account restore is for convenience and after error checking,
calls stream restore, for each stream.

Signed-off-by: Matthias Hanel <mh@synadia.com>

### DIFF
--- a/cli/account_command.go
+++ b/cli/account_command.go
@@ -34,6 +34,10 @@ type actCmd struct {
 	healthCheck       bool
 	snapShotConsumers bool
 	force             bool
+
+	showProgress     bool
+	placementCluster string
+	placementTags    []string
 }
 
 func configureActCommand(app commandHost) {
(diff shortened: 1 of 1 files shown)


cli/account_command.go
```diff
@@ -136,6 +146,39 @@ func (c *actCmd) backupAction(_ *kingpin.ParseContext) error {
...
+		existingStreams[n] = struct{}{}
+	}
+	de, err := os.ReadDir(c.backupDirectory)
+	kingpin.FatalIfError(err, "setup failed")
+	for _, d := range de {
+		if !d.IsDir() {
+			kingpin.FatalIfError(err, "expected a directory")
+		}
+		if _, ok := existingStreams[d.Name()]; ok {
+			kingpin.Fatalf("stream %q exists already", d.Name())
```
ripienaar, line 167: wondering if maybe we should support a replace mode? Otherwise LGTM

ripienaar (approved):

cli/account_command.go
matthiashanel: I am checking these beforehand, because stream restore fails them. Basically want to avoid a partial restore.
So any change would have to start with stream restore.

Would that basically delete the stream before restoring it?

ripienaar: Yes or prompts for replace/skip? With matching flags to say always skip or always replace without prompting for non interactive use. 

I wasn’t sure about this when I made the backup and Todd wanted a release but the choices are basically as above or what you have now as default with flags for always skip and always delete - ie. no prompts.

still not sure tbh

matthiashanel: Since you are not sure, I will skip that for now.

PR #445, opened 2022-05-02
Catches and logs Fetch getting a timeout

jnmoyne:


jnmoyne:
Supersedes #444

cli/bench_command.go
```diff
@@ -786,5 +799,7 @@ func (c *benchCmd) runSubscriber(bm *bench.Benchmark, nc *nats.Conn, startwg *sy
 
 	bm.AddSubSample(bench.NewSample(numMsg, c.msgSize, start, end, nc))
 
+	c.fetchTimeout = fetchTimeout
```
ripienaar, line 802: might need to set some lock on this, also a bit of a race here I think.

If 5 subscribers are running and 1 has a issue, it could set this to false if it exits first and then the others can set it to true so you wont know this.

So maybe we need a lock and a ```if c.fetchTimeout { c.fetchTimeout = fetchTimeout }```?

jnmoyne: fixed c.fetchTimeout now only set if fetchTimeout is true

ripienaar (approved):
LGTM

Issue #465, opened 2022-05-16
Allow customizing the name of the connection via a new command line flag --connection-name


dsidirop:
Essentially what the subject says. It's useful when running certain benchmark tests in rudimentary nats-servers on localdev. In such cases nats-top is employed to generate snapshots of the traffic and it makes sense to make the connection-names distinct as there might be more than one 'nats sub' commands running in parallel.

PR #466, opened 2022-05-16
#465 connection name flag

dsidirop:


nats/main.go
```diff
@@ -44,9 +44,12 @@ See 'nats cheat' for a quick cheatsheet of commands
 	}
 	cli.SetVersion(version)
 
+	defaultConnectionName := "NATS CLI Version " + version
+
 	ncli.Flag("server", "NATS server urls").Short('s').Envar("NATS_URL").PlaceHolder("NATS_URL").StringVar(&opts.Servers)
 	ncli.Flag("user", "Username or Token").Envar("NATS_USER").PlaceHolder("NATS_USER").StringVar(&opts.Username)
 	ncli.Flag("password", "Password").Envar("NATS_PASSWORD").PlaceHolder("NATS_PASSWORD").StringVar(&opts.Password)
+	ncli.Flag("connection-name", "Nickname to use for the underlying NATS Connection (Default is '"+defaultConnectionName+"')").Default(defaultConnectionName).PlaceHolder("NAME").StringVar(&opts.ConnectionName)
```
ripienaar, line 52: I think listing the default here in the output might be a bit long, I think we can just not print the default here - for most people this will never matter

cli/util.go
```diff
@@ -388,7 +388,7 @@ func natsOpts() []nats.Option {
 	kingpin.FatalIfError(err, "configuration error")
 
 	return append(copts, []nats.Option{
-		nats.Name("NATS CLI Version " + Version),
+		nats.Name(opts.ConnectionName),
```
ripienaar, line 391: Above this check if `opts.ConnectionName` is empty and set it to default, that way even when embedded the cli behavior wont change

ripienaar (changes requested):
looks good, minor tweaks :)

nats/main.go
dsidirop: All done.

cli/util.go
dsidirop: Done.

cli/cli.go
```diff
@@ -69,7 +69,9 @@ type Options struct {
 	TlsCA string
 	// Timeout is how long to wait for operations
 	Timeout time.Duration
-	// Username is the username or token to connect with
+	// ConnectionName is the name to use for the underlying NATS connection
+	ConnectionName string
+	// Nkey is the file holding a nkey to connect with
```
ripienaar, line 74: Sorry I didn’t see this before but comments went wrong here. Other than that this is all good!

ripienaar (changes requested):
Comments in cli.go are incorrect - see comment added there

thanks!

ripienaar (approved):

ripienaar:
Thank you

PR #468, opened 2022-05-19
New 'Multi Subjects' flag for all benchmark modes (also: don't delete the consumer if a consumer name is specified, fix consumer name argument name)

jnmoyne:


cli/bench_command.go
```diff
@@ -667,13 +696,13 @@ func (c *benchCmd) runSubscriber(bm *bench.Benchmark, nc *nats.Conn, startwg *sy
 				progress.TimeStarted = startTime
 			}
 			if c.pull {
-				sub, err = js.PullSubscribe(c.subject, c.consumerName)
+				sub, err = js.PullSubscribe("getSubscribeSubject(c)", c.consumerName, nats.BindStream(c.streamName))
```
ripienaar, line 699: probably dont want the quotes here

ripienaar (changes requested):
small tweak only

cli/bench_command.go
jnmoyne: Ah yeah sorry I forgot to clean that up (that parameter is actually completely ignored (yup one of the many reasons we're redoing the JS API :)). Thanks for spotting this.

ripienaar (approved):
LGTM

PR #477, opened 2022-05-30
Add stream number of subject info

ramonberrutti:
Add the number of subjects to the info command.

WDYT to add to report?

### DIFF
--- a/cli/stream_command.go
+++ b/cli/stream_command.go
@@ -1706,6 +1706,10 @@ func (c *streamCmd) showStreamInfo(info *api.StreamInfo) {
 
 	fmt.Printf("     Active Consumers: %d\n", info.State.Consumers)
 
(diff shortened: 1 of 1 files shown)


cli/stream_command.go
```diff
@@ -1705,6 +1705,7 @@ func (c *streamCmd) showStreamInfo(info *api.StreamInfo) {
 	}
 
 	fmt.Printf("     Active Consumers: %d\n", info.State.Consumers)
+	fmt.Printf("   Number of Subjects: %d\n", info.State.NumSubjects)
```
ripienaar, line 1708: I'd wrap this in an if >0 since older servers dont supply this information - and the display is getting giant already

ripienaar (approved):

ripienaar:
Thank you :)

PR #484, opened 2022-06-02
Fixes nats bench multisubject option for batch publications

jnmoyne:
Note, also fixes trying to define the consumers when a consumer name is explicitly passed as an argument


ripienaar (approved):
LGTM

PR #487, opened 2022-06-03
fix #486 - config.StoreSir path escaping on `server run` for windows OS

### LINKED ISSUE #486: `nats server run` fails on windows with `--jetstream`
```
$ nats server run --jetstream
nats.exe: error: Parse error on line 11: 'Invalid escape character 'U'. Only the following escape characters are allowed: \xXX, \t, \n, \r, \", \\.', try --help
```
Apparently the backslash paths of Windows break the parsing at https://github.com/nats-io/natscli/blob/4d32be8d4dddf5b722f0b0a5aad587bab040cd7b/cli/server_run_command.go#L55

(store dir would have been C:**\U**sers\...

AlbericC:
fixes #486

### DIFF
--- a/cli/server_run_command.go
+++ b/cli/server_run_command.go
@@ -8,6 +8,7 @@ import (
 	"os/signal"
 	"os/user"
 	"path/filepath"
+	"runtime"
 	"strings"
 	"syscall"
 	"text/template"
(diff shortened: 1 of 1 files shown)


ripienaar:
thank you

PR #501, opened 2022-06-16
Enabling Bearer Token Authentication in CLI.

shurya-kumar:
This will allow users to authenticate to the NATS servers using a bearer user JWT

### DIFF
--- a/cli/util.go
+++ b/cli/util.go
@@ -446,16 +446,6 @@ func newNatsConnUnlocked(servers string, copts ...nats.Option) (*nats.Conn, erro
 		servers = opts.Config.ServerURL()
 	}
 
-	if opts.UserJwt != "" {
-		copts = append(copts, []nats.Option{
-			nats.UserJWT(func() (string, error) {
-				return opts.UserJwt, nil
-			}, func(sign []byte) ([]byte, error) {
-				return []byte(""), nil
-			}),
-		}...)
-	}
-
 	var err error
 
 	opts.Conn, err = nats.Connect(servers, copts...)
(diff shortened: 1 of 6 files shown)


cli/util.go
```diff
@@ -446,6 +446,16 @@ func newNatsConnUnlocked(servers string, copts ...nats.Option) (*nats.Conn, erro
 		servers = opts.Config.ServerURL()
 	}
 
+	if opts.UserJwt != "" {
```
ripienaar, line 449: Thank you, to be in line with how things work we should probably add this to the natscontext package also so users can save this in their different contextes.  I can do that later this week if its too much for you to take on

shurya-kumar: It needs to be exposed here also rite?

"github.com/nats-io/jsm.go/natscontext"

ripienaar: Yeah that one, so we need a property there and helpers to set it, and helpers to create the nats options once its set.  Then the update here is just adding the flag.

ripienaar: I've merged that, if you rebase here and do `GOPRIVATE=github.com/nats-io go get github.com/nats-io/jsm.go@main` it will update your go.mod with that version and tests ehre will pass

ripienaar: Think you'll first have to rebase onto cli main then do the go get before it will pass

glehmann:
I'd love to have this feature! Is there something blocking this PR?

ripienaar:
It probably just need a rebase on main - github doesnt notify when new pushes are done and @shurya-kumar did not mention it so I probably just missed the update

PR #506, opened 2022-06-17
resolve some confusion in bools

ripienaar:
All booleans had --no-flag versions but often this
just made no sense, setting --no-memory when the default
is false didnt really add any value and just cluttered help
output.

Now we add the nagatable version only where it makes
sense which hopefully remove some confusion


ripienaar:
@jnmoyne I didnt tweak bench command but FIY a few changes in CLI parsing you can now choose between for example `--progress` only or having both `--progress` and `--no-progress`.  The `--help` now shows both version of flags so I think there are a few cases in bench where it would make sense to update some flags.

Issue #509, opened 2022-06-17
consumer sub does not support deliver group


bruth:
The `consumer sub` command does not consider if the consumer has a deliver group.

https://github.com/nats-io/natscli/blob/main/cli/consumer_command.go#L1358

PR #510, opened 2022-06-17
Support deliver group for consumer sub

bruth:
Fixes nats-io/natscli#509

### DIFF
--- a/cli/consumer_command.go
+++ b/cli/consumer_command.go
@@ -195,6 +195,7 @@ func configureConsumerCommand(app commandHost) {
 	consSub.Arg("consumer", "Consumer name").StringVar(&c.consumer)
 	consSub.Flag("ack", "Acknowledge received message").Default("true").BoolVar(&c.ack)
 	consSub.Flag("raw", "Show only the message").Short('r').UnNegatableBoolVar(&c.raw)
+	consSub.Flag("deliver-group", "Deliver group of the consumer").StringVar(&c.deliveryGroup)
 
 	conCluster := cons.Command("cluster", "Manages a clustered Consumer").Alias("c")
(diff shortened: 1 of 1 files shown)


bruth:
:+1: Just opened in response to a user finding it in Slack, so I figured I'd add it

cli/consumer_command.go
```diff
@@ -195,6 +195,7 @@ func configureConsumerCommand(app commandHost) {
 	consSub.Arg("consumer", "Consumer name").StringVar(&c.consumer)
 	consSub.Flag("ack", "Acknowledge received message").Default("true").BoolVar(&c.ack)
 	consSub.Flag("raw", "Show only the message").Short('r').UnNegatableBoolVar(&c.raw)
+	consSub.Flag("deliver-group", "Deliver group of the consumer").StringVar(&c.deliveryGroup)
```
alfrunes, line 198: Isn't the deliver group inferred from the `ConsumerInfo`, which makes this argument unused?

bruth: Do you mean if we fetch the consumer info we could grab this value?

ripienaar: That would be good - but time perhaps better spent on “nats sub” to support binding and behaviours like that via the normal client. 

These consumer sub and pull commands predate us  having clients so I am inclined to deprecate them and keep just nats sub.

alfrunes: > Do you mean if we fetch the consumer info we could grab this value?

Correct me if I'm wrong, but isn't the `ConsumerInfo` [fetched from the API](https://github.com/bruth/natscli/blob/c935f5afbe877e918a95c49e53832bd34befbb9f/cli/consumer_command.go#L1441) already?
If I understand the code correctly the attribute this flag sets is only used in the `create` and `edit` sub-commands, and therefore unused in the `sub` in this context.

PR #511, opened 2022-06-17
Update un-negetable boolean flags for bench

jnmoyne:


ripienaar:
Something else to consider you will see the —help output has changed a lot. You can now have king help text and it will show in bench help not other help. 

Not looked at this PR yet but saying in case your wanted to expand help output. See request help for example

jnmoyne:
Added long help.

Only thing I'm not sure about is whether to keep the progress bar default or change it to be no progress bar by default (progress bar is great to visualize the performance and for first time users, but in many cases it will result in lower numbers) - would you prefer progress bar on or off by default?

cli/bench_command.go
```diff
@@ -69,38 +69,73 @@ const (
 
 func configureBenchCommand(app commandHost) {
 	c := &benchCmd{}
+
+	benchHelp := `
+Core NATS publish and subscribe:
+
+nats bench benchsubject --pub 1 --sub 10
```
ripienaar, line 76: Maybe indent the example commands by 2 or 3 chars so they stand out from the text? Assuming the help parser parses that right

ripienaar (commented):
I like the progress on by default myself

cli/bench_command.go
jnmoyne: Added a couple of spaces, good idea and parser ok with it

ripienaar (approved):
LGTM. Most bools turns out to be these un-negatable ones, but I didnt want to break backwards compat when I added it.  But for sure would be better if bool defaulted to this :(

Do you think the new --help format is better? (nats --help, per command ones not changed much)

jnmoyne:
I agree it would be better if bool was non-neg by default but then you must have false as your default value - thanks for making me think of that as it means I have now cleaned all my default(false) for those flags :).

Yes I think the long help up front will definitely help first time and irregular users as there is so much you can do.

PR #512, opened 2022-06-18
Adds 'lastpersubject' flag to the sub command

jnmoyne:


cli/sub_command.go
```diff
@@ -62,6 +63,7 @@ func configureSubCommand(app commandHost) {
 	act.Flag("new", "Delivers only future messages (requires JetStream)").UnNegatableBoolVar(&c.deliverNew)
 	act.Flag("last", "Delivers the most recent and all future messages (requires JetStream)").UnNegatableBoolVar(&c.deliverLast)
 	act.Flag("since", "Delivers messages received since a duration (requires JetStream)").PlaceHolder("DURATION").StringVar(&c.deliverSince)
+	act.Flag("lastpersubject", "Deliver the most recent messages for each subject in the Stream (requires JetStream)").UnNegatableBoolVar(&c.deliverLastPerSubject)
```
ripienaar, line 66: we have like ```--headers-only``` so while it would be a  bit wordy ```--last-per-subject``` would be more in line with the overall UX here

ripienaar (approved):
LGTM

PR #525, opened 2022-07-08
Rework nats server run

ripienaar:
The behavior of always cleaning contexts after each
run and creating new random passwords and ports on
each start was too difficult to use, especially given
the lack of natscontext support in any other languages.

Now when the required contexts exist the passwords and
port is read from them and reused.  Cleaning behavior
can be enabled with --clean


codegangsta (approved):
Yup. This will definitely work well for the demos. Thanks for getting this in @ripienaar ! 👍

PR #531, opened 2022-07-16
Add the ability to ls keys of a bucket

dselans:
Needed the ability to easily list the keys in a bucket.

Probably doing something against convention - let me know if anything needs to be fixed up.

Thanks!

EDIT: Attached screenshot of what it looks like.
image

### DIFF
--- a/cli/kv_command.go
+++ b/cli/kv_command.go
@@ -14,6 +14,7 @@
 package cli
 
 import (
+	"errors"
 	"fmt"
 	"io"
 	"math"
@@ -31,27 +32,29 @@ import (
 )
 
 type kvCommand struct {
-	bucket              string
-	key                 string
-	val                 string
-	raw                 bool
-	history             uint64
-	ttl                 time.Duration
-	replicas            uint
-	force               bool
-	maxValueSize        int64
-	maxValueSizeString  string
-	maxBucketSize       int64
-	maxBucketSizeString string
-	revision            uint64
-	description         string
-	listNames           bool
-	storage             string
-	placementCluster    string
-	placementTags       []string
-	repubSource         string
-	repubDest           string
-	repubHeadersOnly    bool
+	bucket                string
+	key                   string
+	val                   string
+	raw                   bool
+	history               uint64
+	ttl                   time.Duration
+	replicas              uint
+	force                 bool
+	maxValueSize          int64
+	maxValueSizeString    string
+	maxBucketSize         int64
+	maxBucketSizeString   string
+	revision              uint64
+	description           string
+	listNames             bool
+	lsVerbose             bool
+	lsVerboseDisplayValue bool
+	storage               string
+	placementCluster      string
+	placementTags         []string
+	repubSource           string
+	repubDest             string
+	repubHeadersOnly      bool
 }
 
 func configureKVCommand(app commandHost) {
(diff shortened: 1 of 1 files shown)


cli/kv_command.go
```diff
@@ -122,8 +125,11 @@ NOTE: This is an experimental feature.
 	watch.Arg("bucket", "The bucket to act on").Required().StringVar(&c.bucket)
 	watch.Arg("key", "The key to act on").Default(">").StringVar(&c.key)
 
-	ls := kv.Command("ls", "List available Buckets").Alias("list").Action(c.lsAction)
+	ls := kv.Command("ls", "List available buckets or the keys in a bucket").Alias("list").Action(c.lsAction)
+	ls.Arg("bucket", "The bucket to list the keys").StringVar(&c.bucket)
 	ls.Flag("names", "Show just the bucket names").Short('n').UnNegatableBoolVar(&c.listNames)
+	ls.Flag("verbose", "Show detailed info about the key").Short('v').BoolVar(&c.lsVerbose)
```
ripienaar, line 131: can also be UnNegatableBoolVar, same with --display-value

cli/kv_command.go
```diff
@@ -174,8 +180,95 @@ func (c *kvCommand) lsAction(_ *fisk.ParseContext) error {
...
+		err = c.lsBuckets(mgr)
+	}
+
+	return err
+}
+
+func (c *kvCommand) lsBucketKeys(mgr *jsm.Manager) error {
+	if mgr == nil {
+		return errors.New("mgr cannot be nil")
+	}
```
ripienaar, line 195: as this is only used internall and we know how we call it, and mgr setup failure already handled elsewhere I think this check can go - error would anyway not help users.

And actually mgr isnt used here at all so remove the argument I think, we can then move the call to create the mgr into `lsBuckets` as thats the only place its used

cli/kv_command.go
```diff
@@ -174,8 +180,95 @@ func (c *kvCommand) lsAction(_ *fisk.ParseContext) error {
...
+
+	kv, err := js.KeyValue(c.bucket)
+	if err != nil {
+		return fmt.Errorf("unable to load bucket: %s", err)
+	}
+
+	keys, err := kv.Keys()
+	if err != nil {
+		if err == nats.ErrNoKeysFound {
+			return fmt.Errorf("no keys found in bucket %s", c.bucket)
```
ripienaar, line 211: I dont think this is an error condition, see for example how we handle no streams, we just print a message and clean exit

https://github.com/nats-io/natscli/blob/7b15a8de2c742fac2083759285b59c098ae49ea7/cli/stream_command.go#L2416

ripienaar (changes requested):
some minor changes, thanks for adding this!

cli/kv_command.go
dselans: Sure thing. Will do.

cli/kv_command.go
dselans: 👍

cli/kv_command.go
dselans: Sounds good. Will update.

dselans:
@ripienaar addressed your comments.

cli/kv_command.go
```diff
@@ -176,12 +182,93 @@ func (c *kvCommand) strForOp(op nats.KeyValueOp) string {
...
+
+	kv, err := js.KeyValue(c.bucket)
+	if err != nil {
+		return fmt.Errorf("unable to load bucket: %s", err)
+	}
+
+	keys, err := kv.Keys()
+	if err != nil {
+		if err == nats.ErrNoKeysFound {
+			fmt.Printf("No keys found in bucket")
```
ripienaar, line 206: should be `fmt.Println`

ripienaar (changes requested):
small nit otherwise looks good thank you!

cli/kv_command.go
dselans: Bah! Sorry about that! Fixed.

ripienaar:
Thank you

PR #539, opened 2022-07-24
Feature: Ignored Subjects for Sub Command

boris-ilijic:
## Context 
Wiretapping is a popular pattern also in the phase of developing/debugging an application. In that case, we create NATS  subscription to include all subjects and can preview which messages are published on NATS by various application producers.

## Problem
Sometimes there are too noisy subjects (e.g. pings, heartbeats, alive messages) that frequently produce NATS messages and prevent us from seeing core/useful communication when we have a wiretap subscription.

## Solution 
Provide the way to use wiretap subscription and ignore 'noisy' subjects. So we want to be subscribed to all subjects, but we also want to exclude a few of them just to remove noisy traffic.

### Requirements
Expand `sub` command with `--ignore-subject` flag. Within the flag it is possible to specify:
- One or many subjects
- Subject literal, single or multiple wildcards

### Example
Create `natscli` wiretap subscription with `--ignore-subject` flag:
```
nats sub 'appx.>' --ignore-subject 'appx.beat, appx.alive.*, appx.ping.>'
```
Publish messages which will be caught with the subscription, but some of them will be excluded from the output:
```
nats pub appx.hey.you msg 
nats pub appx.beat msg # will be excluded
nats pub appx.alive.123 msg # will be excluded
nats pub appx.ping.client.123 msg # will be excluded
```

Output:
```
Subscribing on appx.>
Ignored subjects: appx.beat, appx.alive.*, appx.ping.>

[#1] Received on "appx.hey.you"
msg
```

### DIFF
--- a/cli/sub_command.go
+++ b/cli/sub_command.go
@@ -48,7 +48,7 @@ type subCmd struct {
 	headersOnly           bool
 	stream                string
 	jetStream             bool
-	excludeSubjets        []string
+	excludeSubjects       []string
 }
 
 func configureSubCommand(app commandHost) {
@@ -73,7 +73,7 @@ func configureSubCommand(app commandHost) {
 	act.Flag("since", "Delivers messages received since a duration (requires JetStream)").PlaceHolder("DURATION").StringVar(&c.deliverSince)
 	act.Flag("last-per-subject", "Deliver the most recent messages for each subject in the Stream (requires JetStream)").UnNegatableBoolVar(&c.deliverLastPerSubject)
 	act.Flag("stream", "Subscribe to a specific stream (required JetStream)").PlaceHolder("STREAM").StringVar(&c.stream)
-	act.Flag("exclude-subjects", "Subjects for which corresponding messages will be excluded and therefore not shown in the output").StringsVar(&c.excludeSubjets)
+	act.Flag("exclude-subjects", "Subjects for which corresponding messages will be excluded and therefore not shown in the output").StringsVar(&c.excludeSubjects)
 }
 
 func init() {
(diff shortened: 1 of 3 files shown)


ripienaar:
Thanks I will review later this week or next as I am out for a while. Thank you!

boris-ilijic:
Hi @ripienaar,
Do you have any update/opinion about this feature? If you think, the feature can be useful to others, I can fix merge conflicts. In another case, I'll just close PR. Thanks for info.

ripienaar:
Rebase would be good and fix the comment - will then review again closely next week but seems I just had the one comment

boris-ilijic:
Rebase of the branch is done. You mentioned the comment, but unfortunately I cannot see any comment (maybe it's in pending - review hasn't been submitted...)

cli/sub_command.go
```diff
@@ -71,6 +73,7 @@ func configureSubCommand(app commandHost) {
 	act.Flag("since", "Delivers messages received since a duration (requires JetStream)").PlaceHolder("DURATION").StringVar(&c.deliverSince)
 	act.Flag("last-per-subject", "Deliver the most recent messages for each subject in the Stream (requires JetStream)").UnNegatableBoolVar(&c.deliverLastPerSubject)
 	act.Flag("stream", "Subscribe to a specific stream (required JetStream)").PlaceHolder("STREAM").StringVar(&c.stream)
+	act.Flag("exclude-subjects", "Subjects for which corresponding messages will be excluded and therefore not shown in the output").StringsVar(&c.excludeSubjects)
```
ripienaar, line 76: Maybe make this `--ignore-subject` which describes more whats happening and add a `Short('I').PlaceHolder("SUBJECT")`

ripienaar (changes requested):
You're right I reviewed this but never submitted the review, sorry for that.

This is pretty good just some minor comments

cli/sub_command.go
boris-ilijic: yes, much better term.

ripienaar:
thank you!

PR #542, opened 2022-07-26
Match replies to request in `nats sub`

Haramaty:
This adds a `--match-replies` flag to `nats sub` which matches replies to any request sent to the subscribed subject.

The feature is implemented using an additional subscription on `_INBOX.>` subject, and a map matching reply subjects to requests.

### DIFF
--- a/cli/sub_command.go
+++ b/cli/sub_command.go
@@ -35,6 +35,7 @@ type subCmd struct {
 	raw                   bool
 	jsAck                 bool
 	inbox                 bool
+	match                 bool
 	dump                  string
 	limit                 uint
 	sseq                  uint64
@@ -45,6 +46,7 @@ type subCmd struct {
 	deliverLastPerSubject bool
 	headersOnly           bool
 	stream                string
+	jetStream             bool
 }
 
 func configureSubCommand(app commandHost) {
@@ -57,6 +59,7 @@ func configureSubCommand(app commandHost) {
 	act.Flag("durable", "Use a durable consumer (requires JetStream)").StringVar(&c.durable)
 	act.Flag("raw", "Show the raw data received").Short('r').UnNegatableBoolVar(&c.raw)
 	act.Flag("ack", "Acknowledge JetStream message that have the correct metadata").BoolVar(&c.jsAck)
+	act.Flag("match-replies", "Match replies to requests").UnNegatableBoolVar(&c.match)
 	act.Flag("inbox", "Subscribes to a generate inbox").Short('i').UnNegatableBoolVar(&c.inbox)
 	act.Flag("count", "Quit after receiving this many messages").UintVar(&c.limit)
 	act.Flag("dump", "Dump received messages to files, 1 file per message. Specify - for null terminated STDOUT for use with xargs -0").PlaceHolder("DIRECTORY").StringVar(&c.dump)
@@ -74,7 +77,7 @@ func init() {
 	registerCommand("sub", 17, configureSubCommand)
 }
 
-func (c *subCmd) subscribe(_ *fisk.ParseContext) error {
+func (c *subCmd) subscribe(p *fisk.ParseContext) error {
 	nc, err := newNatsConn("", natsOpts()...)
 	if err != nil {
 		return err
@@ -91,12 +94,12 @@ func (c *subCmd) subscribe(_ *fisk.ParseContext) error {
 		return fmt.Errorf("generating inboxes is not compatible with dumping to stdout using null terminated strings")
 	}
 
-	jetStream := c.sseq > 0 || len(c.durable) > 0 || c.deliverAll || c.deliverNew || c.deliverLast || c.deliverSince != "" || c.deliverLastPerSubject || c.stream != ""
+	c.jetStream = c.sseq > 0 || len(c.durable) > 0 || c.deliverAll || c.deliverNew || c.deliverLast || c.deliverSince != "" || c.deliverLastPerSubject || c.stream != ""
 
-	if c.inbox && jetStream {
+	if c.inbox && c.jetStream {
 		return fmt.Errorf("generating inboxes is not compatible with JetStream subscriptions")
 	}
-	if c.queue != "" && jetStream {
+	if c.queue != "" && c.jetStream {
 		return fmt.Errorf("queu group subscriptions are not supported with JetStream")
 	}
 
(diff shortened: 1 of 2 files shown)


cli/sub_command.go
```diff
@@ -139,89 +145,51 @@ func (c *subCmd) subscribe(_ *fisk.ParseContext) error {
...
+		// if reached limit and matched all requests
+		if ctr == c.limit && len(match_map) == 0 {
+			reply_sub.Unsubscribe()
 			cancel()
 		}
 	}
 
+	if c.match {
+		match_map = make(map[string]*nats.Msg)
+		reply_sub, err = nc.Subscribe("_INBOX.>", match_handler)
```
ripienaar, line 184: Please check if `InboxPrefix` is set then use that otherwise `_INBOX`

cli/sub_command.go
```diff
@@ -309,3 +277,116 @@ func (c *subCmd) subscribe(_ *fisk.ParseContext) error {
 
 	return nil
 }
+
+func print_msg(c *subCmd, msg *nats.Msg, reply *nats.Msg, ctr uint) {
```
ripienaar, line 281: Please call this `printMsg` rather is more go idiomatic

cli/sub_command.go
```diff
@@ -309,3 +277,116 @@ func (c *subCmd) subscribe(_ *fisk.ParseContext) error {
...
+			log.Printf("Could not save message: %s", err)
+		}
+
+		if ctr%100 == 0 {
+			fmt.Print(".")
+		}
+	}
+}
+
+func pretty_print(msg *nats.Msg, headersOnly bool) {
```
ripienaar, line 375: Call is `prettyPrintMsg`

cli/sub_command.go
```diff
@@ -309,3 +277,116 @@ func (c *subCmd) subscribe(_ *fisk.ParseContext) error {
...
+			}
+
+			pretty_print(reply, c.headersOnly)
+
+		}
+
+	} // output format type dispatch
+}
+
+func dump_msg(msg *nats.Msg, stdout bool, filepath string, ctr uint) {
```
ripienaar, line 351: call it `dumpMsg`

cli/sub_command.go
```diff
@@ -139,89 +145,51 @@ func (c *subCmd) subscribe(_ *fisk.ParseContext) error {
...
-			sub.Unsubscribe()
+		// if reached limit and matched all requests
+		if ctr == c.limit && len(match_map) == 0 {
+			reply_sub.Unsubscribe()
 			cancel()
 		}
 	}
 
+	if c.match {
+		match_map = make(map[string]*nats.Msg)
```
ripienaar, line 183: all these vars should be like matchMap not match_map

ripienaar (changes requested):
Nice idea I like it but probably it will be for the next CLI release so I will land a few PRs today and then we rebase and review this one thoroughly for the next release.

The print logic has been a area of many bugs so I want to be quite careful when toushing this

Sorry for the delay here I was out for a few weeks

Haramaty:
Hi, thank you for the PR 😊 
Notice I also rebased on top of current main.

Sorry on all the name conventions, I haven't written Go in some years. Let me know when you're reviewing this again, or if you need me to rebase again 👍🏼

ripienaar:
nice, thanks for the update, I will test it during the week carefully and let you know! nice idea thank you

Haramaty:
I forgot to run `go fmt` and failed the lint check, sorry 😬 
Fixed now

cli/sub_command.go
```diff
@@ -139,89 +145,56 @@ func (c *subCmd) subscribe(_ *fisk.ParseContext) error {
...
-				if m.Reply != "" {
-					fmt.Printf("[#%d] Received on %q with reply %q\n", ctr, m.Subject, m.Reply)
-				} else {
-					fmt.Printf("[#%d] Received on %q\n", ctr, m.Subject)
-				}
-			} else if jetStream {
-				fmt.Printf("[#%d] Received JetStream message: stream: %s seq %d / subject: %s / time: %v\n", ctr, info.Stream(), info.StreamSequence(), m.Subject, info.TimeStamp().Format(time.RFC3339))
-			} else {
-				fmt.Printf("[#%d] Received JetStream message: consumer: %s > %s / subject: %s / delivered: %d / consumer seq: %d / stream seq: %d\n", ctr, info.Stream(), info.Consumer(), m.Subject, info.Delivered(), info.ConsumerSequence(), info.StreamSequence())
+		if ctr == c.limit {
```
ripienaar, line 154: I wonder if we should count a matched reply in this counter, it seems weird that when I do `nats sub --match-replies --count 2 '>'` that I get 1 message with its reply.  Seems intuitive that I should get 2 messages and their replies.

What do you think?

ripienaar, line 154: I wonder if we should count a matched reply in this counter, it seems weird that when I do `nats sub --match-replies --count 2 '>'` that I get 1 message with its reply.  Seems intuitive that I should get 2 messages and their replies.

What do you think?

cli/sub_command.go
```diff
@@ -139,89 +145,56 @@ func (c *subCmd) subscribe(_ *fisk.ParseContext) error {
...
+			return
+		}
 
-		} // output format type dispatch
+		printMsg(c, request, reply, ctr)
+		delete(matchMap, reply.Subject)
 
-		if ctr == c.limit {
-			sub.Unsubscribe()
+		// if reached limit and matched all requests
```
ripienaar, line 175: If we do decide to not count these 2 the message count we can just remove this here

ripienaar (commented):
This is great, just a small question about behavior otherwise i think its mergable

cli/sub_command.go
Haramaty: I agree, counting only requests/messages (& ignoring replies) seems more intuitive to us as well. That's the current implementation 😊

image
> on the left: sending two requests. top right: the example reply server. bottom right: a tap matching replies with `--count 2`. The tap closes after two requests (i.e. not counting their replies)

We do count published messages, of course:

image
> on the left: sending one request, and publishing one message. top right: the example reply server. bottom right: same tap from before. Can see one request & its reply, and one published message.

cli/sub_command.go
Haramaty: When we got `count` requests, we still have to wait for a response to any unmatched request

cli/sub_command.go
ripienaar: OK thanks, I tested against the JS API and I think what I saw was just some weirdness cos of how JS works, the code works fine.  Thank you

ripienaar:
Thanks for the contribution!

Issue #546, opened 2022-07-29
Bench with multi-subject not multi publisher aware


derekcollison:
When using multi-subject and bench to stress a KV, if you use multiple publishers they stomp on each other.

e.g. send 1M messages with 20 publishers only gives you 50k keys.

jnmoyne:
--multisubject in the bench command doesn't apply to KV testing, instead you should use the `--kv` flag which works even if you have multiple publisher.

e.g.
```
nats bench testkv --pub 20 --kv --msgs 1000000 --no-progress
12:59:03 KV mode, using the subject name as the KV bucket name. Publishers do puts, subscribers do gets
12:59:03 Starting KV benchmark [bucket=testkv, kv=true, msgs=1,000,000, msgsize=128 B, maxbytes=1.0 GiB, pubs=20, sub=0, storage=memory, replicas=1, pubsleep=0s, subsleep=0s]
12:59:03 Starting KV putter, putting 50,000 messages
etc...

nats kv info testkv
Information for Key-Value Store Bucket testkv created 2022-07-29T12:56:12-07:00

Configuration:

         Bucket Name: testkv
        History Kept: 1
       Values Stored: 1000000
  Backing Store Kind: JetStream
         Description: nats bench bucket
 Maximum Bucket Size: 1073741824
  Maximum Value Size: -1
    JetStream Stream: KV_testkv
             Storage: Memory
```

jnmoyne:
When. you use `--multisubject` right now indeed each publisher has it's own subject generator starting at 0, so are you asking to have all of them to start at a different offset? (however `--multisubjectmax` would still apply 'overall')

PR #547, opened 2022-07-29
Issue 546

jnmoyne:
Addresses https://github.com/nats-io/natscli/issues/546
Publishers publish on a subject with an offset when using --multisubject


derekcollison (approved):
LGTM

PR #551, opened 2022-08-03
Bump nats.go client version to support direct gets by subject

codegangsta:


codegangsta:
Whoops. Need to bump server version as well

jnmoyne (approved):
LGTM

ripienaar:
What would be the CLI benefit other than what is fine in KV? Just checking if that’s all. 


anyway before releasing on Friday I will update all the dependencies etc so generally don’t micro manage these

codegangsta:
@ripienaar Really only to support direct get by subject in KV: https://github.com/nats-io/nats-server/pull/3325

No other reason beyond that

codegangsta:
Relevant client PR: https://github.com/nats-io/nats.go/pull/1030

ripienaar:
Ok, server uodate not required for that. 

Feel free to merge but as mentioned it’s generally not needed to micro manage here next release (Friday if the server is out in time) will pull all latest

codegangsta:
Server update was actually required for this one, in order for the tests to pass. @kozlovic and I chatted on whether or not we should impose a server version check in the client on an iteration of an already very new feature, but decided against it.

codegangsta:
Thanks for the 👀 @ripienaar !

ripienaar:
If tests failed without the server uodate then that needs to be looked at and fixed. 

We can’t have a major CLI feature like KV only work on tbe latest server. 

So I am in favour of reverting this if that is the case.

codegangsta:
Okay. Let me revert this and check out the tests

codegangsta:
I see what you are saying, I think it's just a matter of making the tests a bit clearer. I think technically we are safe because of al l this being gated behind `AllowDirect`, but we can certainly be more cautious

PR #552, opened 2022-08-03
Revert "Bump nats.go client version to support direct gets by subject"

codegangsta:
Reverts nats-io/natscli#551

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -14,10 +14,10 @@ require (
 	github.com/gosuri/uiprogress v0.0.1
 	github.com/guptarohit/asciigraph v0.5.5
 	github.com/kballard/go-shellquote v0.0.0-20180428030007-95032a82bc51
-	github.com/klauspost/compress v1.15.9
+	github.com/klauspost/compress v1.15.7
 	github.com/nats-io/jsm.go v0.0.34-0.20220708145958-5a418c85744e
-	github.com/nats-io/nats-server/v2 v2.8.5-0.20220803150712-d7847c97c116
-	github.com/nats-io/nats.go v1.16.1-0.20220803180602-ec49000ee1e2
+	github.com/nats-io/nats-server/v2 v2.8.5-0.20220708145433-da305f2044cb
+	github.com/nats-io/nats.go v1.16.1-0.20220610202224-dcbb65a13ee9
 	github.com/nats-io/nuid v1.0.1
 	github.com/prometheus/client_golang v1.12.2
 	github.com/prometheus/common v0.35.0
--- a/go.sum
+++ b/go.sum
@@ -173,9 +173,8 @@ github.com/kballard/go-shellquote v0.0.0-20180428030007-95032a82bc51 h1:Z9n2FFNU
 github.com/kballard/go-shellquote v0.0.0-20180428030007-95032a82bc51/go.mod h1:CzGEWj7cYgsdH8dAjBGEr58BoE7ScuLd+fwFZ44+/x8=
 github.com/kisielk/gotool v1.0.0/go.mod h1:XhKaO+MFFWcvkIS/tQcRk01m1F5IRFswLeQ+oQHNcck=
 github.com/klauspost/compress v1.15.5/go.mod h1:PhcZ0MbTNciWF3rruxRgKxI5NkcHHrHUDtV4Yw2GlzU=
+github.com/klauspost/compress v1.15.7 h1:7cgTQxJCU/vy+oP/E3B9RGbQTgbiVzIJWIKOLoAsPok=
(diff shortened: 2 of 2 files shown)


ripienaar:
@kozlovic so is this the intended behaviour of the KV support in nats.go that it only works in latest server?

before reverting and causing more churn let’s just see if we can figure out if it’s expected behaviour.

ripienaar:
As far as I understood it would try to make the bucket with the new setting - this would be ignored by old server. And KV client would not use direct gets if the stream isn’t configured with that option thus old and new streams and servers are supported?

codegangsta:
Correct. KVs currently created in the cli don't do direct gets. So they will still work for previous server versions. However, this will no longer work with an older server if a kv stream is upgraded to `AllowDirect`.

I'm assuming this is indented behavior, but I agree we should check with @kozlovic

kozlovic:
If the KV stream is created with AllowDirect, then the direct get should work with latest of server: note that v2.9.0 has not been released yet, but the "direct" get feature and allow direct will technically be available only in 2.9.0+.

kozlovic:
We may still have a bug, but if KV code uses direct only if stream has the allow_direct boolean set, then it should not be a problem (if the server is latest from main)

ripienaar:
Then as you say against an old server if the KV tried to create it and it didn’t end up getting that setting it should work. 

Perhaps the client assumes because it asks for direct support or will get it and then it goes and just use it without again checking the created bucket? Wild guess. 

We don’t need to revert just check out older commit and see if we can reproduce to see what’s before we decide to revert

codegangsta:
Yup. I'll test a KV with `AllowDirect` on 2.8.4 and verify is still uses non-direct gets

kozlovic:
The client KV code when creating the stream sets the AllowDirect boolean to true and check the returned stream info to have it still set in order to set an internal boolean that dictates if direct-get can be used. Again, for the "bind" to existing KV, the stream info must have AllowDirect set to set the internal boolean.

codegangsta:
So I was able to test this with server `2.8.4` and `2.9.0-beta.18` and found that I could create and operate kvs just fine. In `2.8.4` we can't even set allowdirect since the server doesn't know about it, which makes sense.

The only nit we would run into here is if someone was running nightlies post `AllowDirect` being merged in and was using it. But yeah, no issues between releases on my end.

ripienaar:
OK then we need to see what those failed tests was

codegangsta:
Agreed

ripienaar:
```
+ go test -v --failfast -p=1 ./...
=== RUN   TestCLIKVGet
Error:     kv_test.go:56: nats utility failed: exit status 1
        nats: error: context deadline exceeded
--- FAIL: TestCLIKVGet (5.08s)
```

unfortunately tests dont give much useful there

ripienaar:
If I reproduce that scenario locally and run that test with trace:

```
=== RUN   TestCLIKVGet
    kv_test.go:56: nats utility failed: exit status 1
        21:20:27 >>> $JS.API.STREAM.INFO.KV_T
        
        
        21:20:27 <<< $JS.API.STREAM.INFO.KV_T: {"type":"io.nats.jetstream.api.v1.stream_info_response","config":{"name":"KV_T","subjects":["$KV.T.\u003e"],"retention":"limits","max_consumers":-1,"max_msgs":-1,"max_bytes":-1,"max_age":0,"max_msgs_per_subject":1,"max_msg_size":-1,"discard":"new","storage":"file","num_replicas":1,"duplicate_window":120000000000,"sealed":false,"deny_delete":true,"deny_purge":false,"allow_rollup_hdrs":true,"allow_direct":true},"created":"2022-08-03T19:20:26.773746Z","state":{"messages":1,"bytes":40,"first_seq":1,"first_ts":"2022-08-03T19:20:26.775833Z","last_seq":1,"last_ts":"2022-08-03T19:20:26.775833Z","num_subjects":1,"consumer_count":0}}
        21:20:27 >>> $JS.API.DIRECT.GET.KV_T.$KV.T.X.Y
        
        
        nats: error: context deadline exceeded
        exit status 1
```

codegangsta:
Yup. So it looks like in the test It's calling the API that doesn't exist. Investigating...

ripienaar:
OK, so I reckon this is just something that affects nightlies but it will work fine between releases so we dont need to revert.

Essentially:

 * Old nightly did set the direct supported flag
 * Old nightly did return the supported flag
 * KV tried to use direct feature old nightly didnt have

So its nightly weirdness, we dont need to revert.

codegangsta:
Yes. I suspect that is what is happening, Previous server dependency was pinned to https://github.com/nats-io/nats-server/commit/da305f2044cb, which I believe had direct get support in it. It will time out on that nightly commit. 

I'm going to run the tests against 2.8.4 release and see what happens

codegangsta:
Sounds good. Thanks @ripienaar

PR #553, opened 2022-08-03
Stream forced delete

jnmoyne:
`nats stream rm -f` does a force delete by sending directly the request message rather than trying to get info and a Stream structure from jsm.go first (in case the stream in question is stuck in a bad (e.g. 'offline') state where trying to get the stream's info results in an error.

### DIFF
--- a/cli/stream_command.go
+++ b/cli/stream_command.go
@@ -2326,22 +2326,36 @@ func (c *streamCmd) addAction(pc *fisk.ParseContext) (err error) {
 }
 
 func (c *streamCmd) rmAction(_ *fisk.ParseContext) (err error) {
-	c.connectAndAskStream()
-
 	if !c.force {
+		c.connectAndAskStream()
 		ok, err := askConfirmation(fmt.Sprintf("Really delete Stream %s", c.stream), false)
 		fisk.FatalIfError(err, "could not obtain confirmation")
 
 		if !ok {
 			return nil
 		}
+
+		stream, err := c.loadStream(c.stream)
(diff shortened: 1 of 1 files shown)


cli/stream_command.go
```diff
@@ -2326,22 +2326,36 @@ func (c *streamCmd) addAction(pc *fisk.ParseContext) (err error) {
...
+		fisk.FatalIfError(err, "could not remove Stream")
+	} else { // force delete attempt, even if somehow you can not get info on the stream in question (e.g. it's "offline")
+		c.nc, c.mgr, err = prepareHelper("", natsOpts()...)
+		fisk.FatalIfError(err, "setup failed")
+
+		var resp api.JSApiStreamDeleteResponse
 
-	err = stream.Delete()
-	fisk.FatalIfError(err, "could not remove Stream")
+		msg, err := c.nc.Request(fmt.Sprintf(api.JSApiStreamDeleteT, c.stream), []byte(""), time.Second)
```
ripienaar, line 2349: As per slack, I'd rather not do API calls direct here. Happy to handle this issue next time in the code but the consistent thing to do would be to add a method on the jsm package to do a delete by calling the API, I'd also do that call once the normal route fails

ripienaar: We presumably also need this for consumers since the new server relocate code also leaves orphaned consumers when you try to use that

ripienaar:
Have replaced this with https://github.com/nats-io/natscli/pull/554/commits/cdc79521b751246b962694496b143ae87eb3943c as we discussed on slack

PR #587, opened 2022-09-27
Filter out non-JS enabled servers

bruth:
The report for JetStream summary includes non-JS enabled servers. It is unclear if it should be filtered here or the upstream source of `$SYS.REQ.SERVER.PING.JSZ`, however this appears to be serviced by each server and the CLI is aggregating these responses, correct?

### DIFF
--- a/cli/server_report_command.go
+++ b/cli/server_report_command.go
@@ -127,6 +127,11 @@ func (c *SrvReportCmd) reportJetStream(_ *fisk.ParseContext) error {
 			return err
 		}
 
+		// Ignore non JetStream-enabled servers.
+		if !response.Server.JetStream {
+			continue
+		}
+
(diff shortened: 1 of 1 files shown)


ripienaar:
Not sure, I can see both sides - a server thats supposed to be a JS one but thats gone and disabled jetstream due to a error condition would then just silently vanish. I think it best if it showed up.

So today the only option we have is to show all since there's no way to distinguish those.

bruth:
Interesting, I had not thought about that case. My mental model was more around the raft meta group.. where by these are the servers that can participate in leader election. That circumstance would apply here as well, no? But I guess you are saying that a previously JS-enabled server could be disabled, but still have JS data reported?

ripienaar:
they would respond to jsz yeah I think - and meta group would show them offline presumably

bruth:
That is what I am observing so that would make sense. Ok, will close since its a valid circumstance.

Issue #593, opened 2022-09-30
Progress bar on nats object put|get work inverted


rickardgranberg:
Was doing some testing of the Object Store feature and didn't understand why I didn't see the progress bars.

I ran: `nats object put <bucket> <file>` and `nats object put --progress <bucket> <file>` expecting to see a progress bar but did not. And yes, my file was over 20480 bytes.

To my surprise, running `nats object put --no-progress <bucket> <file>` makes the progress bar appear!

I've traced it to the lines https://github.com/nats-io/natscli/blob/1057211892d576f6b46c8c3a4861abb457e5668f/cli/object_command.go#L85 https://github.com/nats-io/natscli/blob/1057211892d576f6b46c8c3a4861abb457e5668f/cli/object_command.go#L97

The flag is named `--progress` with a default of `true` but the variable it's assigned to is named `noProgress` and it's used in that capacity on line https://github.com/nats-io/natscli/blob/1057211892d576f6b46c8c3a4861abb457e5668f/cli/object_command.go#L486 and https://github.com/nats-io/natscli/blob/1057211892d576f6b46c8c3a4861abb457e5668f/cli/object_command.go#L562

I'll be happy to provide a PR with a fix, but I'm a little uncertain if the intention was to have progress enabled or disabled by default.

ripienaar:
hmmm, yeah i think logic is just reversed, should be always on

rickardgranberg:
Alright. I'll work on a fix

PR #594, opened 2022-09-30
Fix progress bars not working

rickardgranberg:
This is a suggested fix for issue #593 

I decided to rename the flag to `--no-progress` to match the one in the `bench` command

### DIFF
--- a/cli/object_command.go
+++ b/cli/object_command.go
@@ -38,7 +38,7 @@ type objCommand struct {
 	overrideName        string
 	hdrs                []string
 	force               bool
-	noProgress          bool
+	progress            bool
 	storage             string
 	listNames           bool
 	placementCluster    string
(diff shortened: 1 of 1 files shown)


cli/object_command.go
```diff
@@ -82,7 +82,7 @@ NOTE: This is an experimental feature.
 	put.Flag("name", "Override the name supplied to the object store").StringVar(&c.overrideName)
 	put.Flag("description", "Sets an optional description for the object").StringVar(&c.description)
 	put.Flag("header", "Adds headers to the object").Short('H').StringsVar(&c.hdrs)
-	put.Flag("progress", "Disables progress bars").Default("true").BoolVar(&c.noProgress)
+	put.Flag("no-progress", "Disable progress bars").UnNegatableBoolVar(&c.noProgress)
```
ripienaar, line 85: since the new cli library now have negatable booleans I think I'd rather like:

```
put.Flag("progress", "Disable progress bars").Default("true").BoolVar(&c.progress)
```

like in stream backup and restore, its a bit more in tune also with how we handle other boolean flags now

rickardgranberg: OK. Update coming!

PR #600, opened 2022-10-05
Allow to not unsubscribe when using sub

sylr:
Signed-off-by: Sylvain Rabot <sylvain@abstraction.fr>

### DIFF
--- a/cli/sub_command.go
+++ b/cli/sub_command.go
@@ -18,9 +18,11 @@ import (
 	"encoding/json"
 	"fmt"
 	"os"
+	"os/signal"
 	"path/filepath"
 	"strings"
 	"sync"
+	"syscall"
 	"time"
 
 	"github.com/choria-io/fisk"
@@ -48,6 +50,7 @@ type subCmd struct {
 	headersOnly           bool
 	stream                string
 	jetStream             bool
+	unsubscribe           bool
 	ignoreSubjects        []string
 }
 
@@ -74,6 +77,7 @@ func configureSubCommand(app commandHost) {
 	act.Flag("last-per-subject", "Deliver the most recent messages for each subject in the Stream (requires JetStream)").UnNegatableBoolVar(&c.deliverLastPerSubject)
 	act.Flag("stream", "Subscribe to a specific stream (required JetStream)").PlaceHolder("STREAM").StringVar(&c.stream)
 	act.Flag("ignore-subject", "Subjects for which corresponding messages will be ignored and therefore not shown in the output").Short('I').PlaceHolder("SUBJECT").StringsVar(&c.ignoreSubjects)
+	act.Flag("unsubscribe", "Unsubscribe on termination").Default("true").BoolVar(&c.unsubscribe)
 }
 
 func init() {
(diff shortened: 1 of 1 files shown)


cli/sub_command.go
```diff
@@ -74,6 +77,7 @@ func configureSubCommand(app commandHost) {
 	act.Flag("last-per-subject", "Deliver the most recent messages for each subject in the Stream (requires JetStream)").UnNegatableBoolVar(&c.deliverLastPerSubject)
 	act.Flag("stream", "Subscribe to a specific stream (required JetStream)").PlaceHolder("STREAM").StringVar(&c.stream)
 	act.Flag("ignore-subject", "Subjects for which corresponding messages will be ignored and therefore not shown in the output").Short('I').PlaceHolder("SUBJECT").StringsVar(&c.ignoreSubjects)
+	act.Flag("unsubscribe", "Unsubscribe on termination").Default("true").BoolVar(&c.unsubscribe)
```
ripienaar, line 80: Wouldnt it be better to say lets just always unsubscribe UNLESS we were given a consumer name to bind to? 

Anyway unsubscribe seems not needed since we alway smake ephemeral consumers or was this somehow causing a problem?

sylr: The purpose of this is to test an edge case of JetStream.

If you subscribe with a consumer which does not exist and unsubscribe within the same "session" then the consumer is deleted.

However, If you did not unsubscribe during the session that created the consumer (let's say a panic occurred) the consumer remains. Upon re-connection, if you do perform an unsubscribe the consumer is not deleted because the consumer has not been created in this session.

ripienaar: yes, thats (unfortunately) how it works, however since `nats sub` is making ephemerals (unless bind is given) the consumers are going away anyway on their own.

So safest thing for the command line is just not to unsubscribe as we cant anticipate all cases where people are using it and unsubscribe is inherently unsafe

sylr: The purpose of this is not to make this cli safe, it is to make it possible to replicate the different behaviours you could hit based on how you use Unsubscribe().

ripienaar: I fundementally think this is the wrong place to introduce those kinds of abilities.  I see the utility of that, but its something we should rather add to some form of testing tool, perhaps `nats bench`

sylr: I don't see why, this just allows you to replicate what would happen if you do not Unsubscribe() after creating a consumer with Subscribe, something that can very well happen and is completely related to the `sub` command.

If you prefer, we can turn it around and call this `no-unsubscribe` (make it false by default) and explain that this is for generating an abandoned consumer.

ripienaar: Because I prefer the core tools be tools that work in most cases and that cant accidentically cause harm to users systems.

We could totally design and build a tool to test streange situations in here, but I dont think `nats sub` is for that. If you want to open a issue with the kinds of scenarios you'd like to test in this manner I will give it some thought how such a tool might look

sylr: But the `sub` command can create abandoned consumers.

Do a `killall -9 nats` or cut the WiFi while subscribing and there you go. This option just allows to do that in a controlled manner to understand the implications.

To me this tools serves as a way to discover NATS and hiding edge cases like this one helps no one.

ripienaar: They are epehemeral and the server cleans them up.  They will not linger.

sylr: Unless you are using a JS Durable consumer.

ripienaar: Sorry, not really sure what to say here, this is not a bug testing / causing harness, I suggested I would create such a thing for you if you open an issue.  I'll close this now.

PR #602, opened 2022-10-05
[FIXED] Discard Policy was incorrectly displayed.

kozlovic:
For instance, a discard "new" policy was displayed as:
```
Discard Policy: NewNew
```

Signed-off-by: Ivan Kozlovic <ivan@synadia.com>

### DIFF
--- a/cli/stream_command.go
+++ b/cli/stream_command.go
@@ -1611,7 +1611,7 @@ func (c *streamCmd) showStreamConfig(cfg api.StreamConfig) {
 	if cfg.DiscardNewPer {
 		dnp = "New Per Subject"
 	}
-	fmt.Printf("       Discard Policy: %s%s\n", cfg.Discard.String(), dnp)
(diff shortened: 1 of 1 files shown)


ripienaar (approved):
LGTM thank you

kozlovic:
@ripienaar Not sure if there are tests for the UI, if there are, please point them to me and I can try to add a test for that.

ripienaar:
There are tests but more testing the side effect based stuff. Say you create a stream with a option I then check tbe option is in effect. 

For the pure display stuff no tests 

see cli package there is a test there but I think fine to skip here

ripienaar:
Sorry I mean https://github.com/nats-io/natscli/blob/main/nats/nats_test.go

kozlovic:
Yes, I saw that test, but indeed more testing that the resulting state is matching the params, etc.., but not UI per-se. Do you merge or am I supposed to?

ripienaar:
You can merge

PR #612, opened 2022-10-27
add {{BODY}} token to access request body within --command

miraculli:
I add a new {{BODY}} token to access the request body within --command 

Additionally I wrote a line about the ENVs that get attached to the spawns cmd.

### DIFF
--- a/cli/reply_command.go
+++ b/cli/reply_command.go
@@ -52,6 +52,14 @@ This will request the weather for london when invoked as:
 
   nats request weather.london ''
 
+Use {{BODY}} to access the request body within the --command
+  
+The command gets also spawned with two ENVs:
+  NATS_REQUEST_SUBJECT
+  NATS_REQUEST_BODY
+
+  nats reply 'echo' --command="printenv NATS_REQUEST_BODY" 
+  
 The body and Header values of the messages may use Go templates to create unique messages.
 
    nats reply test "Message {{Count}} @ {{Time}}"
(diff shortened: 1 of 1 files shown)


ripienaar:
Is this different from the existing "{{.Request}}" support?

ripienaar:
See here https://github.com/nats-io/natscli/blob/3e7b7bef0e4871feea5f8ad55cab3c20fbd29dd4/cli/util.go#L602-L639

ripienaar:
```
$ nats reply x 'received: {{.Request}}'
$ nats req x 'hello world'
received: hello world
```

miraculli:
ah... didn´t know about the `{{.Request}}`
Sorry for the noise.

ripienaar:
All good, if you want you can still add the comments update?

miraculli:
Good morning, 
I just checked `{{.Request}}` but it seems to be different to what I hand in mind with this PR.
The `{{.Request}}` variable doesn't get resolved inside a --command exec like

```
nats reply 'test' --command="echo {{.Request}}"
```


```
nats request test 'heyya'
09:38:57 Sending request on "test"
09:38:57 Received with rtt 43.645593ms
{{.Request}}
```

The reply command should exec a `echo heyya` in this case and the response should be just `heyya`

ripienaar:
Ok. I will fix that. Better it works the same as other places rather then. Will check today

ripienaar:
This should sort it https://github.com/nats-io/natscli/pull/613

PR #652, opened 2022-12-15
SOCKS Proxy support for contexts and NATS

philpennock:


philpennock:
Depends upon https://github.com/nats-io/jsm.go/pull/428

philpennock:
The logic when I first pushed this worked if the proxy were set in the context file, which is how I'd tested, but not if set via the command-line or environ.  The updates I've pushed fix that to actually work "normally".

ripienaar (approved):

ripienaar:
thank you :)

PR #655, opened 2022-12-20
Display cluster size in "nats server report jetstream"

rodrigc:
Add a new table "RAFT Meta Cluster Information" to display this.

Signed-off-by: Craig Rodrigues <rodrigc@crodrigues.org>

Inspired by discussion here:
https://natsio.slack.com/archives/CM3T6T7JQ/p1671467758952119?thread_ts=1671465211.070579&cid=CM3T6T7JQ

### DIFF
--- a/cli/server_report_command.go
+++ b/cli/server_report_command.go
@@ -267,6 +267,13 @@ func (c *SrvReportCmd) reportJetStream(_ *fisk.ParseContext) error {
 			cNames = names
 		}
 
+		metaClusterTable := newTableWriter("RAFT Meta Cluster Information")
+		metaClusterTable.AddHeaders("Name", "Leader", "Peer", "Size")
+		metaClusterTable.AddRow(cluster.Name, cluster.Leader, cluster.Peer, cluster.Size)
+
+		fmt.Print(metaClusterTable.Render())
+		fmt.Println()
+
 		table := newTableWriter("RAFT Meta Group Information")
 		table.AddHeaders("Name", "ID", "Leader", "Current", "Online", "Active", "Lag")
 		for i, replica := range cluster.Replicas {


jnmoyne:
Thanks for this. Definitely shouldn’t have to do --trace to get the cluster size! Looking at your thread and how you had to work your way through understanding how to resolve the issue, I would even go further and add a message that says explicitly (when currently you don’t have enough nodes up to reach a meta quorum) that you need to either restart one of ‘list of ids’ currently missing nodes or add a new one to restore service.

rodrigc:
Thanks for the feedback.

Yes, analyzing and root causing the problem was challenging.  @derekcollison was very helpful with that.

One other idea I have is that instead of printing out a message in the `nats server report jetstream`, instead do:

1.  If  `nats server report jetstream` shows that there is no master node (cluster has not reached quorum)
2.  CLI can print out a URL where user can go read and find more detailed info about how to fix the problem.

The reason I suggest a URL, is because I don't have a good feel for how to word a message which is sensible, and doesn't clutter the UI of `nats server report jetstream`

ripienaar:
With this change the output now looks like here:

Screenshot 2023-01-06 at 12 23 48

This seems redundant the only new information is that there are 9 servers.  Something we could highlight in the heading of the existing table perhaps?

Screenshot 2023-01-06 at 12 28 12

This is from the slack thread, the down peers were already shown so I am not sure this additional table showing same information would have helped? 

Imo we should keep the report as is and add the warning line @jnmoyne suggests, these 2 tables would not clarify that

ripienaar:
Further if we recreate the problem where there are too many machines down and no leader, then these tables dont show up because in order for them to be rendered there has to be a leader.

```
$ docker compose ps
NAME                COMMAND                  SERVICE              STATUS              PORTS
nc1-c1              "/nats-server --conf…"   nc1.c1.example.net   running             6222/tcp, 8222/tcp, 0.0.0.0:10000->4222/tcp, :::10000->4222/tcp
nc1-c2              "/nats-server --conf…"   nc1.c2.example.net   exited (137)
nc2-c1              "/nats-server --conf…"   nc2.c1.example.net   running             6222/tcp, 8222/tcp, 0.0.0.0:10001->4222/tcp, :::10001->4222/tcp
nc2-c2              "/nats-server --conf…"   nc2.c2.example.net   exited (137)
nc3-c1              "/nats-server --conf…"   nc3.c1.example.net   running             6222/tcp, 8222/tcp, 0.0.0.0:10002->4222/tcp, :::10002->4222/tcp
nc3-c2              "/nats-server --conf…"   nc3.c2.example.net   exited (137)
```

Here I have a 2 x 3 super cluster with one cluster down, there's no leader because its impossible.

Now neither the existing or new meta cluster table will be shown since it requires a meta leader.

Would you be ok if I replace this PR with one that does this in that scenario:

```
╭───────────────────────────────────────────────────────────────────────────────────────────────╮
│                                       JetStream Summary                                       │
├────────┬─────────┬─────────┬───────────┬──────────┬───────┬────────┬──────┬─────────┬─────────┤
│ Server │ Cluster │ Streams │ Consumers │ Messages │ Bytes │ Memory │ File │ API Req │ API Err │
├────────┼─────────┼─────────┼───────────┼──────────┼───────┼────────┼──────┼─────────┼─────────┤
│ nc1-c1 │ c1      │ 0       │ 0         │ 0        │ 0 B   │ 0 B    │ 0 B  │ 0       │ 0       │
│ nc2-c1 │ c1      │ 0       │ 0         │ 0        │ 0 B   │ 0 B    │ 0 B  │ 0       │ 0       │
│ nc3-c1 │ c1      │ 0       │ 0         │ 0        │ 0 B   │ 0 B    │ 0 B  │ 0       │ 0       │
├────────┼─────────┼─────────┼───────────┼──────────┼───────┼────────┼──────┼─────────┼─────────┤
│        │         │ 0       │ 0         │ 0        │ 0 B   │ 0 B    │ 0 B  │ 0       │ 0       │
╰────────┴─────────┴─────────┴───────────┴──────────┴───────┴────────┴──────┴─────────┴─────────╯


WARNING: No cluster meta leader found. The cluster expects 6 nodes but only 3 responded. JetStream operation require at least 4 up nodes.
```

rodrigc:
> 
> WARNING: No cluster meta leader found. The cluster expects 6 nodes but only 3 responded. JetStream operation require at least 4 up nodes.
> ```
> 
> If that's fine we can replace this one with #668


OK.

ripienaar:
Replaced by https://github.com/nats-io/natscli/pull/668

PR #665, opened 2023-01-02
Adds '--term' to 'consumer next'

jnmoyne:
Allows you to terminate a message


cli/consumer_command.go
```diff
@@ -191,6 +192,7 @@ func configureConsumerCommand(app commandHost) {
 	consNext.Arg("stream", "Stream name").Required().StringVar(&c.stream)
 	consNext.Arg("consumer", "Consumer name").Required().StringVar(&c.consumer)
 	consNext.Flag("ack", "Acknowledge received message").Default("true").BoolVar(&c.ack)
+	consNext.Flag("term", "Terms the message").Default("false").BoolVar(&c.term)
```
ripienaar, line 195: Should not be negatable otherwise lgtm

jnmoyne: Ah yes forgot that thanks

PR #670, opened 2023-01-09
b60afb7c introduced nil deref when no JS for account

philpennock:


cli/account_command.go
```diff
@@ -502,10 +502,12 @@ func (c *actCmd) infoAction(_ *fisk.ParseContext) error {
 
 	info, err := mgr.JetStreamAccountInfo()
 
-	if info.Domain == "" {
-		fmt.Println("JetStream Account Information:")
-	} else {
-		fmt.Printf("JetStream Account Information for domain %s:\n", info.Domain)
+	if err == nil {
```
ripienaar, line 505: sweet thanks, please drop the new line above and I would rather check if `info != nil` since who knows the `JetStreamAccountInfo()` might have some bug where err is nil but info is also nil.

It's unlikely of course, but since we dont use the err anyway we might as well check the thing we actually care for.

ripienaar (changes requested):
LGTM just a nit on the approach

cli/account_command.go
philpennock: Done.  I think my brain was stuck on trying to merge with the dispatch flow below and when I shrugged and settled for just adjusting it in-place, I didn't switch err->info.

Note that there's one change in behavior from before b60afb7c, in that the section header won't be printed.

ripienaar (approved):
LGTM

PR #673, opened 2023-01-11
Add support for `nats server request profile`

neilalexander:
This adds support for `nats server request profile`, as per the new `PROFILEZ` endpoint in nats-io/nats-server#3774. The profiles are written to files in the supplied directory, so that they can be used with `go tool pprof`.

### DIFF
--- a/cli/server_request_command.go
+++ b/cli/server_request_command.go
@@ -14,7 +14,11 @@
 package cli
 
 import (
+	"encoding/json"
 	"fmt"
+	"os"
+	"path/filepath"
+	"time"
 
 	"github.com/choria-io/fisk"
 	"github.com/nats-io/nats-server/v2/server"
@@ -50,6 +54,10 @@ type SrvRequestCmd struct {
 
 	jsServerOnly bool
 	jsEnabled    bool
+
+	profileName  string
+	profileDebug int
+	profileDir   string
 }
 
 func configureServerRequestCommand(srv *fisk.CmdClause) {
@@ -113,6 +121,11 @@ func configureServerRequestCommand(srv *fisk.CmdClause) {
 	healthz := req.Command("jetstream-health", "Request JetStream health status").Alias("healthz").Action(c.healthz)
 	healthz.Flag("js-enabled", "Checks that JetStream should be enabled on all servers").Short('J').BoolVar(&c.jsEnabled)
 	healthz.Flag("server-only", "Restricts the health check to the JetStream server only, do not check streams and consumers").Short('S').BoolVar(&c.jsServerOnly)
+
+	profilez := req.Command("profile", "Run a profile").Action(c.profilez)
+	profilez.Arg("profile", "Specify the name of the profile to run (allocs, heap, goroutine, mutex, threadcreate, block)").StringVar(&c.profileName)
+	profilez.Arg("dir", "Set the output directory for profile files").Default(".").ExistingDirVar(&c.profileDir)
+	profilez.Flag("debug", "Set the debug level of the profile").IntVar(&c.profileDebug)
 }
 
 func (c *SrvRequestCmd) healthz(_ *fisk.ParseContext) error {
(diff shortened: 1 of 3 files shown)


cli/server_request_command.go
```diff
@@ -113,6 +117,10 @@ func configureServerRequestCommand(srv *fisk.CmdClause) {
 	healthz := req.Command("jetstream-health", "Request JetStream health status").Alias("healthz").Action(c.healthz)
 	healthz.Flag("js-enabled", "Checks that JetStream should be enabled on all servers").Short('J').BoolVar(&c.jsEnabled)
 	healthz.Flag("server-only", "Restricts the health check to the JetStream server only, do not check streams and consumers").Short('S').BoolVar(&c.jsServerOnly)
+
+	profilez := req.Command("profile", "Run a profile").Action(c.profilez)
+	profilez.Flag("profile", "Specify the name of the profile to run").StringVar(&c.profileName)
```
ripienaar, line 122: since its required and has no default we could make this an `Arg()`  and then have `EnumVar(&c.profileName, "cpu", "mem", ....)` so we'll get validtion etc.  In the help portiion here I tend to list the valid enum values in the help string like `"The profile name to request (mem, cpu, ....)"`

ripienaar (changes requested):
small tweaks otherwise lgtm

cli/server_request_command.go
```diff
@@ -138,6 +146,45 @@ func (c *SrvRequestCmd) healthz(_ *fisk.ParseContext) error {
...
+		b = append(b, r...)
+	}
+	if err := json.Unmarshal(b, &response); err != nil {
+		return fmt.Errorf("no response")
+	}
+
+	if response.Resp.Error != "" {
+		return fmt.Errorf(response.Resp.Error)
+	}
+	fmt.Print(string(response.Resp.Profile))
```
ripienaar, line 183: probably need a new line here

ripienaar: Are all the profiles text? nothing thats the pprof gzip stuff?

cli/server_request_command.go
neilalexander: Switching to `Arg()` definitely makes sense. I did wonder about validating the profile name clientside but then if we change profiles in the server later, that would also require a NATS CLI update. Passing through the name and letting the server decide if it's acceptable probably makes more sense here.

cli/server_request_command.go
neilalexander: Since the response is just `[]byte` it did do the right thing for gzipped profiles, but to make it clearer that it might be binary, I've replaced it by writing directly to `os.Stdout` using a buffer.

cli/server_request_command.go
ripienaar: I tend to favor user friendlyness here over the work to keep it friendly :) But yeah this is niche enough and these `nats server req` stuff are advanced enough usage that I think its ok without the enum.

ripienaar: But we should show what is valid in --help at least, even if we dont validate so that as we add more peole arent stuck that they cant access new ones if we tell them in support - and then presumably support will result in a fix issue here

cli/server_request_command.go
```diff
@@ -138,6 +148,49 @@ func (c *SrvRequestCmd) healthz(_ *fisk.ParseContext) error {
...
+	}
+
+	if response.Resp.Error != "" {
+		return fmt.Errorf(response.Resp.Error)
+	}
+
+	// The response might be in binary format (some profiles can return
+	// gzipped responses) so write it to stdout as-is.
+	bio := bytes.NewBuffer(response.Resp.Profile)
+	bio.WriteTo(os.Stdout)
```
ripienaar, line 189: We definitely cant write binary to stdout.

These req commands will fetch from multiple servers and the output would be unusable I think?

Maybe better to force a directory name be given and we output the results into files like `YYYMMDD-HHMMSS-<server id>-<profile name>` that way at least the data can be used and the utility is useful for running in a for loop to do ongoing profile gathering?

neilalexander: Ah, didn't occur to me that multiple servers could respond. I will come up with something like you have suggested.

ripienaar: We could also force a single server only on this one? Still binary to stdout is really bad UX

cli/server_request_command.go
```diff
@@ -113,6 +121,11 @@ func configureServerRequestCommand(srv *fisk.CmdClause) {
 	healthz := req.Command("jetstream-health", "Request JetStream health status").Alias("healthz").Action(c.healthz)
 	healthz.Flag("js-enabled", "Checks that JetStream should be enabled on all servers").Short('J').BoolVar(&c.jsEnabled)
 	healthz.Flag("server-only", "Restricts the health check to the JetStream server only, do not check streams and consumers").Short('S').BoolVar(&c.jsServerOnly)
+
+	profilez := req.Command("profile", "Run a profile").Action(c.profilez)
+	profilez.Arg("profile", "Specify the name of the profile to run (i.e. allocs, heap, goroutine...)").StringVar(&c.profileName)
```
ripienaar, line 126: lets list all the ones we know about today so `Name of the profile to run (allocs, heap, goroutine)` just expand the list to include what we have.  Users would have no way to figure this out.

cli/server_request_command.go
```diff
@@ -113,6 +121,11 @@ func configureServerRequestCommand(srv *fisk.CmdClause) {
...
+	profilez.Arg("dir", "Set the output directory for profile files").Default(".").StringVar(&c.profileDir)
```
ripienaar, line 127: ```ExistingDirVar(&c.profileDir)```

ripienaar (changes requested):

ripienaar:
Looks good, I am seeing build/test failures:

```
Error: ../cli/account_command.go:393:17: undefined: server.UserInfo
Error: ../cli/account_command.go:405:20: undefined: server.UserInfo
```

fix those, squash commits and you can merge when green

cli/server_request_command.go
```diff
@@ -113,6 +121,11 @@ func configureServerRequestCommand(srv *fisk.CmdClause) {
 	healthz := req.Command("jetstream-health", "Request JetStream health status").Alias("healthz").Action(c.healthz)
 	healthz.Flag("js-enabled", "Checks that JetStream should be enabled on all servers").Short('J').BoolVar(&c.jsEnabled)
 	healthz.Flag("server-only", "Restricts the health check to the JetStream server only, do not check streams and consumers").Short('S').BoolVar(&c.jsServerOnly)
+
+	profilez := req.Command("profile", "Run a profile").Action(c.profilez)
+	profilez.Arg("profile", "Specify the name of the profile to run (allocs, heap, goroutine, mutex, threadcreate, block)").StringVar(&c.profileName)
+	profilez.Arg("dir", "Set the output directory for profile files").Default(".").ExistingDirVar(&c.profileDir)
+	profilez.Flag("debug", "Set the debug level of the profile").IntVar(&c.profileDebug)
```
ripienaar, line 128: this one *might* be problematic as we need to be quite careful with duplicate flags and this is very common - but lets leave as is and i will maybe update this before release

PR #674, opened 2023-01-12
stream edit: support removing Placement constraints

philpennock:
* remove a duplicated nil-handling block
* ~~document why we need to use a sentinel~~
* ~~document that we chose 'none' as our sentinel~~
* use fooSet variables to distinguish between "set by user to empty string" and "not set at all"
* use an empty cluster or a single empty tag to remove those constraints

~~This does mean that if someone creates a cluster called `none` or uses `none` as a tag, then they're out of luck with these flags.  We can pick another sentinel value if we think that's a blocker, I'm not invested in `none`, it simply seemed the least bad way out.~~

### DIFF
--- a/cli/stream_command.go
+++ b/cli/stream_command.go
@@ -86,6 +86,8 @@ type streamCmd struct {
 	replicas              int64
 	placementCluster      string
 	placementTags         []string
+	placementClusterSet   bool
+	placementTagsSet      bool
 	peerName              string
 	sources               []string
 	mirror                string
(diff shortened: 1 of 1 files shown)


kettlewell (approved):

philpennock:
Thanks.  Since this is an aesthetics of UX update for a public tool, I'll hold off waiting on an approval from @ripienaar too.

ripienaar:
Will review tomorrow thanks

cli/stream_command.go
```diff
@@ -1267,16 +1267,24 @@ func (c *streamCmd) copyAndEditStream(cfg api.StreamConfig, pc *fisk.ParseContex
 		cfg.Placement = &api.Placement{}
 	}
 
-	if cfg.Placement == nil {
-		cfg.Placement = &api.Placement{}
-	}
+	// We don't have a good way to signal "flag specified as explicitly empty" vs
+	// "flag not specified".  So we have to use sentinels.
+	// We use 'none' because - is dropped by the CLI lib before even reaching us.
```
ripienaar, line 1272: This whole stream edit model is deeply annoying and flawed needs a full redo so yeah, agree this is probably the least worse.  

What we can do is this:

```
f.Flag("discard-per-subject", "Sets the 'new' discard policy and applies it to every subject in the stream").IsSetByUser(&c.discardPerSubjSet).BoolVar(&c.discardPerSubj)
```

Here it would enable you to do `--placement-cluster=""` and you can determine the difference between default empty and specifically set empty using the `IsSetByUser()` variable.  In time as I get motivation I will probably add a  bunch of that for a hopefully less annoying outcome.

If we will treat `none` as magic we should call that out in the `--help` for this flag.

philpennock: I prefer the ability to sense an explicit empty string and not have magic sentinels which squat in the same namespace, documented or otherwise, so will rework this.

cli/stream_command.go
```diff
@@ -1267,16 +1269,22 @@ func (c *streamCmd) copyAndEditStream(cfg api.StreamConfig, pc *fisk.ParseContex
...
+	// command-line" from "is not empty".
 
-	if c.placementCluster != "" {
+	if c.placementClusterSet {
 		cfg.Placement.Cluster = c.placementCluster
 	}
 
-	if len(c.placementTags) > 0 {
-		cfg.Placement.Tags = c.placementTags
+	if c.placementTagsSet {
```
ripienaar, line 1280: nice this turned out much better.  The `IsSetByUser()` behavior is quite new - after I wrote this code all, so nice to see it help.

Will review closely tomorrow, thank you

ripienaar (approved):
nice, LGTM

PR #682, opened 2023-01-25
Rollback context if edit invalid

sethjback:
Since the `cli` launches an external editor validation of the updated context cannot be done until after it has already been saved to disk. The result is when you enter bad data (e.g. invalid `nsc` path) it will get persisted before the cli notifies you of the error. If you then try to run the edit command again it will fail due to validation checks failing when attempting to load the now bad context. The only solution is to manually edit the context `.json` file on disk.

This PR will check for errors when loading the context after the editor command has returned. If the context fails to validate it will attempt to save the known clean context so that the `nats edit context` command will run in the future. It then returns the error to let the user know what part of the updated context failed validation.

This is not a perfect solution: if you have made multiple changes to the context config they all get reverted instead of just the invalid ones. However, this will at least solve the frustrating situation of making a mistake in the context edit and then getting "locked out" of editing it again.

### DIFF
--- a/cli/context_command.go
+++ b/cli/context_command.go
@@ -174,7 +174,15 @@ func (c *ctxCommand) editCommand(pc *fisk.ParseContext) error {
 		return err
 	}
 
-	return c.showCommand(pc)
+	// There was an error with some data in the modified config
+	// Save the known clean version and show the error
+	err = c.showCommand(pc)
+	if err != nil {
+		ctx.Save(c.name)
+		return err
+	}
+
+	return nil
 }
 
 func (c *ctxCommand) renderListCompletion(current string, known []*natscontext.Context) {
--- a/cli/context_command.go
+++ b/cli/context_command.go
@@ -179,7 +179,7 @@ func (c *ctxCommand) editCommand(pc *fisk.ParseContext) error {
 	err = c.showCommand(pc)
 	if err != nil {
 		ctx.Save(c.name)
-		return err
+		return fmt.Errorf("updated context validation failed - rolling back changes\n%w", err)
 	}
 
 	return nil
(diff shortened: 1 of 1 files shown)


cli/context_command.go
```diff
@@ -174,7 +174,15 @@ func (c *ctxCommand) editCommand(pc *fisk.ParseContext) error {
 		return err
 	}
 
-	return c.showCommand(pc)
+	// There was an error with some data in the modified config
+	// Save the known clean version and show the error
+	err = c.showCommand(pc)
+	if err != nil {
+		ctx.Save(c.name)
+		return err
```
ripienaar, line 182: Looks great, lets just expand on this error a tad saying whats going on.

Tbh this should probably edit a temporary copy rather than the actual path to be safe safe butt his is a good improvement already that I am happy to merge as is.

sethjback: Sounds good. I like the idea of editing a temp copy, and we should probably do validation on all the fields as well.

This was just a short path to avoid getting locked out of editing a context via the cli.

cli/context_command.go
```diff
@@ -174,7 +174,15 @@ func (c *ctxCommand) editCommand(pc *fisk.ParseContext) error {
 		return err
 	}
 
-	return c.showCommand(pc)
+	// There was an error with some data in the modified config
+	// Save the known clean version and show the error
+	err = c.showCommand(pc)
+	if err != nil {
+		ctx.Save(c.name)
+		return fmt.Errorf("updated context validation failed - rolling back changes\n%w", err)
```
ripienaar, line 182: cool, sorry to be annoying lets just do go standard way so `"updated context validation failed - rolling back changes: %w"` then its all good to do, do that wait for green and you can merge when ready (if you can, dont know what permission you have yet)

sethjback: Yep, no worries - I added the newline for readability since there is a lot of output embedded in that error return.

ripienaar: hmm, ok, then perhaps better as it was lol - I will play with it and tweak the UX after no worries, thanks for the fix!

PR #694, opened 2023-02-06
Remove spurious print statement

bruth:
This is causing JSON-rendered output to be invalid.

Signed-off-by: Byron Ruth <byron@nats.io>

### DIFF
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -5,7 +5,7 @@ jobs:
   test:
     strategy:
       matrix:
-        go: [ 1.18, 1.19 ]
+        go: [ 1.19, 1.20 ]
 
     env:
       GOPATH: /home/runner/work/natscli
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -5,7 +5,7 @@ jobs:
   test:
     strategy:
       matrix:
-        go: [ 1.19, 1.20 ]
+        go: [ "1.19", "1.20" ]
 
     env:
       GOPATH: /home/runner/work/natscli
--- a/cli/cli.go
+++ b/cli/cli.go
@@ -17,7 +17,6 @@ import (
 	"context"
 	"embed"
 	glog "log"
-	"math/rand"
 	"sort"
 	"sync"
 	"time"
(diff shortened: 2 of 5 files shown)


bruth:
The failure appears to be an install while installing staticcheck.. although I was able to install the same version locally, so I am not sure why it is failing in CI.

ripienaar:
Yeah unfortunately staticcheck has stopped supporting go 1.18, you can remove 1.18 from the github action and it wil go away

bruth:
Does an older version of staticcheck support Go 1.18? I suppose we could use that version for that specific build?

ripienaar:
Happy to drop go 1.18, its not supported by go team anymore, dropping it is easiest :)

.github/workflows/test.yaml
```diff
@@ -5,7 +5,7 @@ jobs:
   test:
     strategy:
       matrix:
-        go: [ 1.18, 1.19 ]
+        go: [ 1.19, 1.20 ]
```
ripienaar, line 8: you'll have to quote `"1.20"`, easiest to just quote both to avoid the mistake in future (1.20 is 1.2 not 1.20)

bruth: :man_facepalming: Yep.

ripienaar: Made the same mistake on other projects hehe

ripienaar:
```
 (220, 2)  SA1019  rand.Seed has been deprecated since Go 1.20 and an alternative has been available since Go 1.0: Programs that call Seed and then expect a specific sequence of results from the global random source (using functions such as Int) can be broken when a dependency changes how much it consumes from the global random source. To avoid such breakages, programs that need a specific result sequence should use NewRand(NewSource(seed)) to obtain a random generator that other packages cannot access. 
```

You can fix it, or just drop 1.20 for now - this is why I suggested not adding 1.20 yet as I know there are a few changes I wanted to go through first

ripienaar (approved):
LGTM

Ignore the test requirement, I already updated github but it wont relfect on this PR

bruth:
:+1: Sounds good. I cannot merge, so feel free.

PR #695, opened 2023-02-06
Support for new stream subject transform functionalities

jnmoyne:
Support for the new stream subject transform functionalities:

- subject tranform for stream sources
- subject tranform for the stream input
- sourcing for KV buckets


cli/server_mapping_command.go
```diff
@@ -57,7 +57,7 @@ func (c *SrvMappingCmd) mappingAction(_ *fisk.ParseContext) error {
 		}
 	}
 
-	trans, err := server.NewSubjectTransformer(c.src, c.dest)
+	trans, err := server.NewSubjectTransform(c.src, c.dest)
```
ripienaar, line 60: Ugh, missed fixing this one in the server, thanks

ripienaar (changes requested):
Let some small comments, I need to see how this looks to really review the UX changes, will look fridayish.

cli/stream_command.go
```diff
@@ -199,6 +201,8 @@ func configureStreamCommand(app commandHost) {
 	strAdd.Flag("validate", "Only validates the configuration against the official Schema").UnNegatableBoolVar(&c.validateOnly)
 	strAdd.Flag("output", "Save configuration instead of creating").PlaceHolder("FILE").StringVar(&c.outFile)
 	addCreateFlags(strAdd, false)
+	strAdd.Flag("subject-transform-source", "Stream subject transform source").PlaceHolder("SUBJECTTRANSFORMSOURCE").StringVar(&c.subjectTransformSource)
+	strAdd.Flag("subject-transform-destination", "Stream subject transform destination").PlaceHolder("SUBJECTTRANSFORMDEST").StringVar(&c.subjectTransformDest)
```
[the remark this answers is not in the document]
jnmoyne: Done

cli/stream_command.go
```diff
@@ -1477,90 +1485,97 @@ func (c *streamCmd) cpAction(pc *fisk.ParseContext) error {
...
 	if len(cfg.Subjects) > 0 {
-		fmt.Printf("             Subjects: %s\n", strings.Join(cfg.Subjects, ", "))
+		fmt.Printf("              Subjects: %s\n", strings.Join(cfg.Subjects, ", "))
+	}
+	if cfg.SubjectTransform != nil && cfg.SubjectTransform.Destination != "" {
+		source := cfg.SubjectTransform.Source
+		if source == "" {
+			source = ">"
+		}
+		fmt.Printf("Subject Transform Source: %s , Destination: %s\n", source, cfg.SubjectTransform.Destination)
```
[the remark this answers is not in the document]
jnmoyne: Indents adjusted

cli/stream_command.go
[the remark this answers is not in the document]
jnmoyne: Yeah I had fixed that already

cli/kv_command.go
```diff
@@ -463,6 +465,17 @@ func (c *kvCommand) addAction(_ *fisk.ParseContext) error {
 		}
 	}
 
+	if len(c.sources) != 0 {
+		var sources []*nats.StreamSource
+
+		for _, source := range c.sources {
+			sources = append(sources, &nats.StreamSource{
+				Name: source,
```
ripienaar, line 473: Would this not require people to enter KV_X as the source name in --source?

cli/kv_command.go
```diff
@@ -787,6 +800,15 @@ func (c *kvCommand) showStatus(store nats.KeyValue) error {
 
 			fmt.Println()
 		}
+
+		if len(nfo.Sources) > 0 {
+			fmt.Println("\n Sources Information:")
+			fmt.Println()
+			for _, source := range nfo.Sources {
+				fmt.Printf("        Source Bucket: %s\n", strings.TrimPrefix(source.Name, "KV_"))
```
ripienaar, line 808: Think it would need to be like stream info - showing lag and last seen also per source

cli/stream_command.go
```diff
@@ -1315,8 +1325,8 @@ func (c *streamCmd) copyAndEditStream(cfg api.StreamConfig, pc *fisk.ParseContex
 		cfg.Placement = nil
 	}
 
-	if len(c.sources) > 0 || c.mirror != "" {
-		return cfg, fmt.Errorf("cannot edit mirrors or sources using the CLI, use --config instead")
+	if len(c.sources) > 0 || c.mirror != "" || c.subjectTransformSource != "" || c.subjectTransformDest != "" {
+		return cfg, fmt.Errorf("cannot edit mirrors, sources or transforms using the CLI, use --config instead")
```
ripienaar, line 1329: We should be able to allow editing igest transform though using flags right?

cli/stream_command.go
```diff
@@ -2170,14 +2212,21 @@ func (c *streamCmd) prepareConfig(_ *fisk.ParseContext, requireSize bool) api.St
...
 		cfg.RePublish = &api.RePublish{
 			Source:      c.repubSource,
 			Destination: c.repubDest,
 			HeadersOnly: c.repubHeadersOnly,
 		}
 	}
 
+	if c.subjectTransformDest != "" {
+		cfg.SubjectTransform = &api.SubjectTransformConfig{
+			Source:      c.subjectTransformSource,
```
ripienaar, line 2225: Do we support empty string source? Seems we should check both are not empty. I know you check it in other places to detect both arent empty but seems kind of action at a distance while it would more maintainable if we check things right here instead

ripienaar (changes requested):
some small comments, need more time to run it and see how the UX feels

ripienaar:
If you update to latest fisk release I think it will now do some dynamic column width adjustment and on larger displays no more wrapping of flags such that we have 10 more chars to play with for long flags.

Update to that lets see how it looks

cli/kv_command.go
jnmoyne: No you don't have to and it's ok if you do, I made the nats.go code handle both cases and use the right stream name either way: https://github.com/nats-io/nats.go/blob/7971018b6c2f0f8b1da88336d415e47940c49ede/kv.go#L418

cli/stream_command.go
jnmoyne: I had it at first but then I realized the problem is that there is no way to use the flags in edit to remove an existing subject transform so I made it that you have to use a config file like source and mirror, and even when I did allow it you must specify both source and dest when doing an edit because otherwise for example edit with just specifying a new dest but not specifying source does it mean remove the existing source or leave it there if there's one?

TLDR just use a config file to make those kinds of stream changes, anyways changing things like replicas or tags I that you may need to do by hand it's good to be able to do with just arguments but IMHO editing a JSON should be the preferred way to change a stream config (though you do have to know the JSON schema).

cli/stream_command.go
```diff
@@ -1409,6 +1419,10 @@ func (c *streamCmd) interactiveEdit(cfg api.StreamConfig) (api.StreamConfig, err
 }
 
 func (c *streamCmd) editAction(pc *fisk.ParseContext) error {
+	if (c.repubSource != "" && c.repubDest == "") || (c.repubSource == "" && c.repubDest != "") {
+		fisk.Fatalf("must specify both --republish-source and --republish-destination")
```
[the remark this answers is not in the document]
jnmoyne: The problem is this:
https://github.com/nats-io/natscli/blob/84138d64116854a7c954fbb0d9e451657d2d7107/cli/stream_command.go#L1354

I discovered during my testing that if you specify just one of the two, it ignores it (and the user gets a 'configs are the same') (same thing with headers-only BTW). So I just added this guard up front while I was there :).

Ultimately, it's similar as above with subject transform source and dest how to distinguish the missing one of the two as meaning set the existing one to "" or as don't modify the existing one. We could go with the meaning a re-use the current value if not specified, but in the end, to me, I still have the same problem as above with not being able to use edit with command line arguments to remove things - it feels that for anything that you may want to add or remove (rather than just change like a replication degree) using command line arguments feels broken.

Like, how to you remove a tag besides editing the JSON?

cli/stream_command.go
```diff
@@ -1679,20 +1706,21 @@ func (c *streamCmd) showStreamInfo(info *api.StreamInfo) {
 		return
 	}
 
-	if c.showStateOnly {
-		fmt.Printf("State for Stream %s created %s\n", c.stream, info.Created.Local().Format("2006-01-02 15:04:05"))
-		fmt.Println()
-	} else {
+	if !c.showStateOnly {
```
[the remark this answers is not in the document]
jnmoyne: Yeah definitely, somehow got switched around as part of the merging changes I had to do after the rebase, thanks for the catch.

Yes I now see what you mean with at least 2 indents. Done.

cli/stream_command.go
jnmoyne: So I made the server transform code tolerant to "" source being treated like ">" source (anywhere transform is used), obviously doesn't make sense to have an empty transform so that's an error returned by the server. So right now it lets you just specify a destination when creating the stream.

That said I'd rather people did always use ">" rather than empty, which I did later by forcing the user to specify both source and dest arguments. And also I made info display a ">" if the config source is "".

It related to the requirement above of requiring both source and dest for the republish.

I can update the if to check both source && dest if you want?

cli/stream_command.go
ripienaar: I added some stuff to the cli library that you can now tell the difference between `--foo=""` and the option not given.  So then this can be made editable by flags. 

https://github.com/nats-io/natscli/blob/84138d64116854a7c954fbb0d9e451657d2d7107/cli/stream_command.go#L171

See the `IsSetByUser()` option there, you check the variable that is set in that way and then you know if the intention of the user is to remove/unset something by specifically setting it to empty.

I havnt yet made the change to do it everywhere it makes sense but have been using it for new things

cli/stream_command.go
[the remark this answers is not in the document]
ripienaar: Same as above I guess with the `IsSetByUser()` then.

cli/stream_command.go
ripienaar: Yeah I am not in favor of the magical empty here, rather people are explicit.

cli/kv_command.go
jnmoyne: Indeed, done.

cli/stream_command.go
jnmoyne: Done. Also added a `--no-transform` flag since republish had one already (and it makes for an easier to find way to remove the transform since most people won't realize it can be done by setting source and dest to "")

cli/stream_command.go
jnmoyne: Done: changed to && on both source and dest

cli/stream_command.go
```diff
@@ -1409,6 +1471,10 @@ func (c *streamCmd) interactiveEdit(cfg api.StreamConfig) (api.StreamConfig, err
 }
 
 func (c *streamCmd) editAction(pc *fisk.ParseContext) error {
+	//if (c.repubSource != "" && c.repubDest == "") || (c.repubSource == "" && c.repubDest != "") {
```
ripienaar, line 1474: remove these lines pls

cli/stream_command.go
```diff
@@ -1719,28 +1800,42 @@ func (c *streamCmd) showStreamInfo(info *api.StreamInfo) {
...
 	}
 
 	showSource := func(s *api.StreamSourceInfo) {
-		fmt.Printf("          Stream Name: %s\n", s.Name)
-		fmt.Printf("                  Lag: %s\n", humanize.Comma(int64(s.Lag)))
+		fmt.Printf("               Stream Name: %s\n", s.Name)
+
+		if s.SubjectTransformDest != "" {
+			if s.FilterSubject == "" {
+				fmt.Printf("            Subject Filter: %s\n", ">")
```
ripienaar, line 1814: `Subject` here is also redundant can just be `Filter` and `Transform`.  Actually I just dont think we need to show it here.

```
Replication:

                   Sources: ORDERS_1, Subject filter: js.in.orders_1, Subject transform: js.in.orders_X

Source Information:

               Stream Name: ORDERS_1
            Subject Filter: js.in.orders_1
         Subject Transform: js.in.orders_X
                       Lag: 0
                 Last Seen: 388ms
```

The first part is the configuration showing filters and such, the 2nd part is the state of the configured replication.

You can see in `nats s state` only the 2nd half is shown, we dont show the config in there so I would remove the `Subject Filter` and `Subject Transform` completely from `Source Information` (and perhaps rename `Source Information` to `Source State`)

cli/stream_command.go
```diff
@@ -2366,6 +2474,10 @@ func (c *streamCmd) validateCfg(cfg *api.StreamConfig) (bool, []byte, []string,
 }
 
 func (c *streamCmd) addAction(pc *fisk.ParseContext) (err error) {
+	if (c.subjectTransformSource != "" && c.subjectTransformDest == "") || (c.subjectTransformSource == "" && c.subjectTransformDest != "") {
```
ripienaar, line 2477: similar comment here as with the one that was in `editAction`, its action at a distance and would be clearer if added to `prepareConfig` I think

ripienaar (changes requested):
Very close now, this is amazing stuff really love it

cli/stream_command.go
```diff
@@ -2282,6 +2384,12 @@ func (c *streamCmd) askSource(name string, prefix string) *api.StreamSource {
 	}, &cfg.FilterSubject)
 	fisk.FatalIfError(err, "could not request filter")
 
+	err = askOne(&survey.Input{
+		Message: fmt.Sprintf("%s Subject mapping transform", prefix),
+		Help:    "Map matching subjects according to this transform destination",
+	}, &cfg.SubjectTransformDest)
+	fisk.FatalIfError(err, "could not request subject mapping destination transform")
```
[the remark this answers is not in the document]
jnmoyne: Yes source already had SubjectFilter, so to introduce transform there it makes sense to re-use the source's subject filter as the source for the transform, but obviously can't rename it to something other because that would break compatibility.

I think the better option here would be to not prompt for the transform (i.e. the destination) if the user didn't specify a filter

cli/stream_command.go
```diff
@@ -1517,90 +1583,97 @@ func (c *streamCmd) cpAction(pc *fisk.ParseContext) error {
...
 	if len(cfg.Subjects) > 0 {
-		fmt.Printf("             Subjects: %s\n", strings.Join(cfg.Subjects, ", "))
+		fmt.Printf("                  Subjects: %s\n", strings.Join(cfg.Subjects, ", "))
+	}
+	if cfg.SubjectTransform != nil && cfg.SubjectTransform.Destination != "" {
+		source := cfg.SubjectTransform.Source
+		if source == "" {
+			source = ">"
+		}
+		fmt.Printf("  Subject Transform Source: %s , Destination: %s\n", source, cfg.SubjectTransform.Destination)
```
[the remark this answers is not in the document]
jnmoyne: Done. And also moved republish info under Options as well

cli/stream_command.go
```diff
@@ -1647,7 +1721,13 @@ func (c *streamCmd) renderSource(s *api.StreamSource) string {
 		parts = append(parts, fmt.Sprintf("Start Time: %v", s.OptStartTime))
 	}
 	if s.FilterSubject != "" {
-		parts = append(parts, fmt.Sprintf("Subject: %s", s.FilterSubject))
+		parts = append(parts, fmt.Sprintf("Subject filter: %s", s.FilterSubject))
```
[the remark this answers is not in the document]
jnmoyne: ok, done

cli/stream_command.go
jnmoyne: Had forgotten to clean them up. Done.

cli/stream_command.go
```diff
@@ -1096,7 +1109,7 @@ func (c *streamCmd) reportAction(_ *fisk.ParseContext) error {
 
 func (c *streamCmd) renderReplication(stats []streamStat) {
 	table := newTableWriter("Replication Report")
-	table.AddHeaders("Stream", "Kind", "API Prefix", "Source Stream", "Active", "Lag", "Error")
+	table.AddHeaders("Stream", "Kind", "API Prefix", "Source Stream", "Filter", "Destination", "Active", "Lag", "Error")
```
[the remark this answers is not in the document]
jnmoyne: Ok with that. Done.

cli/stream_command.go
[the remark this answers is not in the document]
jnmoyne: Done: don't prompt for destination transform if filter subject is ""

cli/stream_command.go
jnmoyne: Fine with changing to just `Filter` and `Transform` however I think I would like to still have the filter and transform repeated in Source Information (now source State) this is because you can now source more than once from the same stream and in that case there will be two (or more) entries in Source Information with the same "Stream Name" and therefore it's nice to have the filter/transform info repeated there so you can tell which source information relates to which subject filter. So leaving the repeat in for the moment, let me know if you're ok with that or not.

ripienaar: Yeah thats fair, so I wonder how we can communicate this without making this block bigger as a stream with many sources could already be pages full.

Maybe something like `Stream: ORDERS_1 Filter: x Transform: y` leaving lag and seen as is today? Kind of small improvement so might not be worth doing so I am ok with this either way, lets just drop the redundant `Subject`

cli/kv_command.go
```diff
@@ -738,8 +751,8 @@ func (c *kvCommand) showStatus(store nats.KeyValue) error {
 	cols.AddSectionTitle("Configuration")
 
 	cols.AddRow("Bucket Name", status.Bucket())
-	cols.AddRow("History Kept", status.History())
-	cols.AddRow("Values Stored", status.Values())
+	cols.AddRow("History Kept", humanize.Comma(status.History()))
```
ripienaar, line 754: dont need to `humanize.Comma`, `AddRow()` does it for you

cli/kv_command.go
```diff
@@ -759,7 +772,7 @@ func (c *kvCommand) showStatus(store nats.KeyValue) error {
 		if nfo.Config.MaxAge <= 0 {
 			cols.AddRow("Maximum Age", "unlimited")
 		} else {
-			cols.AddRow("Maximum Age", nfo.Config.MaxAge)
+			cols.AddRow("Maximum Age", humanizeDuration(nfo.Config.MaxAge))
```
ripienaar, line 775: `AddRow()` also does this for you

ripienaar (changes requested):
Looking good, just remove all the redundant humanize calls.

Only time we need them now is with `AddRowf()`

ripienaar:
@jnmoyne how are we on this? Can you rebase? People desperate for the CLI release so I want to get this one in.

jnmoyne:
I'd also like to get this one finally merged :). Actually rebased yet again 5 days ago but yeah needed another one since, which is done now.

jnmoyne:
The status that it's still to go. Since that PR was created there's been additional functionality (i.e. multi subject support) in stream sources and mirroring which this doesn't cover as I'd rather create a new PR for those.

ripienaar:
> The status is that it's still good to go. Since that PR was created there's been additional functionality (i.e. multi subject support) in stream sources and mirroring which this doesn't cover as I'd rather create a new PR for those. If you'd prefer I did those changes in new commits to this branch instead let me know.

Unfortunately looks like latest clean up introduced a number of formatting bugs and code layout changes etc, left some comments there.

Is the server work in 2.10 now done and merged completely?

I prefer smaller PRs for sure. so do not want to add to this.  Even this I'd have done as many PRs rather than a huge singe PR :)

ripienaar (changes requested):

cli/consumer_command.go
```diff
@@ -572,7 +572,6 @@ func (c *consumerCmd) showInfo(config api.ConsumerConfig, state api.ConsumerInfo
 	cols.AddRowIfNotEmpty("Name", config.Name)
 	cols.AddRowIf("Durable Name", config.Durable, config.Durable != "" && config.Durable != config.Name)
 	cols.AddRowIfNotEmpty("Description", config.Description)
-
```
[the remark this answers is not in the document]
jnmoyne: Good practice indeed I shall remember from now on, done.

cli/kv_command.go
```diff
@@ -744,16 +757,16 @@ func (c *kvCommand) showStatus(store nats.KeyValue) error {
 	if nfo != nil {
 		cols.AddRowIfNotEmpty("Description", nfo.Config.Description)
 
-		cols.AddRow("Bucket Size", humanize.IBytes(nfo.State.Bytes))
+		cols.AddRow("Bucket Size", nfo.State.Bytes)
 		if nfo.Config.MaxBytes == -1 {
 			cols.AddRow("Maximum Bucket Size", "unlimited")
 		} else {
-			cols.AddRow("Maximum Bucket Size", humanize.IBytes(uint64(nfo.Config.MaxBytes)))
```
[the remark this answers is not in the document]
jnmoyne: That's true good reminder that magical function F can't tell if it's just a number of bytes, done.

cli/kv_command.go
```diff
@@ -464,6 +466,17 @@ func (c *kvCommand) addAction(_ *fisk.ParseContext) error {
 		}
 	}
 
+	if len(c.sources) != 0 {
+		var sources []*nats.StreamSource
```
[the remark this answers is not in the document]
jnmoyne: You are correct, actually it doesn't need the if statement either, removed both.

cli/stream_command.go
```diff
@@ -1403,18 +1415,61 @@ func (c *streamCmd) copyAndEditStream(cfg api.StreamConfig, pc *fisk.ParseContex
...
-		cfg.RePublish = &api.RePublish{
-			Source:      c.repubSource,
-			Destination: c.repubDest,
-			HeadersOnly: c.repubHeadersOnly,
-		}
+	var repubConfig api.RePublish
+
+	if cfg.RePublish != nil {
+		repubConfig = *cfg.RePublish
+	} else {
```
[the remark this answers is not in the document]
jnmoyne: Cleaned up

cli/stream_command.go
```diff
@@ -1403,18 +1415,61 @@ func (c *streamCmd) copyAndEditStream(cfg api.StreamConfig, pc *fisk.ParseContex
...
 
 	if c.noRepub {
 		cfg.RePublish = nil
 	}
 
+	var subjectTransformConfig api.SubjectTransformConfig
+
+	if cfg.SubjectTransform != nil {
+		subjectTransformConfig = *cfg.SubjectTransform
+	} else {
```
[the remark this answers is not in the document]
jnmoyne: Removed

cli/stream_command.go
```diff
@@ -1403,18 +1415,61 @@ func (c *streamCmd) copyAndEditStream(cfg api.StreamConfig, pc *fisk.ParseContex
...
+		subjectTransformConfig.Destination = c.subjectTransformDest
+	}
+
+	if subjectTransformConfig.Source == "" && subjectTransformConfig.Destination == "" {
+		cfg.SubjectTransform = nil
+	} else {
+		cfg.SubjectTransform = &subjectTransformConfig
+	}
+
+	if c.noSubjectTransform {
```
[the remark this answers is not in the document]
jnmoyne: You are right that some of it is not needed anymore after I added noSubjectTransform, which is so that you can remove a mapping using stream edit. Originally it tried to assume stream edit w/out specifying a transform source or dest would mean 'remove the transform', but you can't use the absence of the argument to assume you want to not have any filter (incompatible with multiple filters/transforms anyways) hence now there's noSubjectTransform added to edit. Unnecessary code removed.

cli/stream_command.go
```diff
@@ -1775,6 +1848,15 @@ func (c *streamCmd) showStreamInfo(info *api.StreamInfo) {
 
 	showSource := func(s *api.StreamSourceInfo) {
 		cols.AddRow("Stream Name", s.Name)
+		if s.SubjectTransformDest != "" {
+			var filter = ">"
```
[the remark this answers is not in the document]
jnmoyne: Got it and cleaned up at this and other places. Apologies my code may contain leftover reflexes from using a functional language with `val` and `var` (rather than `const`) and bleeding into my Go code where I use `var` to signify 'mutable' which is obviously lost in the translation :).

cli/stream_command.go
```diff
@@ -1816,13 +1899,11 @@ func (c *streamCmd) showStreamInfo(info *api.StreamInfo) {
 	} else {
 		cols.AddRowf("First Sequence", "%s @ %s UTC", f(info.State.FirstSeq), f(info.State.FirstTime))
 	}
-
```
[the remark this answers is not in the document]
jnmoyne: Made the styling consistent (lines between the control statements like if and switch) as best as I could throughout the whole `stream_command.go` file including other parts of the file not related to this PR. It does improve readability.

cli/stream_command.go
```diff
@@ -2224,14 +2303,24 @@ func (c *streamCmd) prepareConfig(_ *fisk.ParseContext, requireSize bool) api.St
 		}
 	}
 
-	if c.repubSource != "" && c.repubDest != "" {
+	if c.repubSource != "" || c.repubDest != "" {
```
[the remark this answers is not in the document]
jnmoyne: The if statement is not needed, cleaned up.

cli/stream_command.go
```diff
@@ -2420,6 +2517,7 @@ func (c *streamCmd) validateCfg(cfg *api.StreamConfig) (bool, []byte, []string,
 }
 
 func (c *streamCmd) addAction(pc *fisk.ParseContext) (err error) {
+
```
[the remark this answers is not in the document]
jnmoyne: Cleaned up.

cli/kv_command.go
```diff
@@ -744,7 +752,7 @@ func (c *kvCommand) showStatus(store nats.KeyValue) error {
 	if nfo != nil {
 		cols.AddRowIfNotEmpty("Description", nfo.Config.Description)
 
-		cols.AddRow("Bucket Size", humanize.IBytes(nfo.State.Bytes))
```
ripienaar, line 747: one was missed

cli/stream_command.go
```diff
@@ -200,12 +203,15 @@ func configureStreamCommand(app commandHost) {
...
 		f.Flag("allow-mirror-direct", "Allows fast, direct, access to stream data via the direct get API on mirrors").IsSetByUser(&c.allowMirrorDirectSet).BoolVar(&c.allowMirrorDirect)
+		f.Flag("transform-source", "Stream subject transform source").PlaceHolder("SOURCE").StringVar(&c.subjectTransformSource)
+		f.Flag("transform-destination", "Stream subject transform destination").PlaceHolder("DEST").StringVar(&c.subjectTransformDest)
 		f.Flag("metadata", "Adds metadata to the stream").PlaceHolder("META").IsSetByUser(&c.metadataIsSet).StringMapVar(&c.metadata)
 		f.Flag("republish-source", "Republish messages to --republish-destination").PlaceHolder("SOURCE").StringVar(&c.repubSource)
 		f.Flag("republish-destination", "Republish destination for messages in --republish-source").PlaceHolder("DEST").StringVar(&c.repubDest)
-		f.Flag("republish-headers", "Republish only message headers, no bodies").UnNegatableBoolVar(&c.repubHeadersOnly)
+		f.Flag("republish-headers", "Republish only message headers, no bodies").BoolVar(&c.repubHeadersOnly)
```
ripienaar, line 211: This doesnt make sense with default being false sice essentially it implies `--no-republish-headers` is the default behavior and so passing that is a noop.

Why change this away from UnNegatable?

cli/stream_command.go
```diff
@@ -1403,18 +1419,53 @@ func (c *streamCmd) copyAndEditStream(cfg api.StreamConfig, pc *fisk.ParseContex
...
+	}
+
+	if c.repubHeadersOnly {
+		repubConfig.HeadersOnly = true
+	} else if repubConfig.HeadersOnly {
+		repubConfig.HeadersOnly = false
+	}
+
+	if (repubConfig.Source == "" || repubConfig.Source == ">") && repubConfig.Destination == "" {
+		cfg.RePublish = nil
```
ripienaar, line 1443: still weird to me to make all the repubConfig stuff and then not use it later on. Ditto later with SubjectTransform

Both would be easier with a func helper so you can just return early from the func in this case

cli/stream_command.go
```diff
@@ -1636,7 +1697,7 @@ func (c *streamCmd) showStreamConfig(cols *columns.Writer, cfg api.StreamConfig)
 	if cfg.MaxBytes == -1 {
 		cols.AddRow("Maximum Bytes", "unlimited")
 	} else {
-		cols.AddRow("Maximum Bytes", humanize.IBytes(uint64(cfg.MaxBytes)))
```
ripienaar, line 1639: missed one

cli/stream_command.go
```diff
@@ -1646,7 +1707,7 @@ func (c *streamCmd) showStreamConfig(cols *columns.Writer, cfg api.StreamConfig)
 	if cfg.MaxMsgSize == -1 {
 		cols.AddRow("Maximum Message Size", "unlimited")
 	} else {
-		cols.AddRow("Maximum Message Size", humanize.IBytes(uint64(cfg.MaxMsgSize)))
```
ripienaar, line 1649: missed one

ripienaar (changes requested):
small changes left

cli/kv_command.go
jnmoyne: Done

cli/stream_command.go
jnmoyne: Done

cli/stream_command.go
jnmoyne: Done

cli/stream_command.go
jnmoyne: That was leftover from when we didn't have --no-republish yet (for steam edit). Cleaned up, simplified and fixed config checking for republish.

cli/stream_command.go
jnmoyne: Yes, that was convoluted and had some unneeded checks, and --no-republish/--no-transform had been tacked on at the end, rewrote for both it's much cleaner now thanks for that.

cli/stream_command.go
ripienaar: think this one still needs its humanize back

cli/stream_command.go
```diff
@@ -2224,6 +2312,10 @@ func (c *streamCmd) prepareConfig(_ *fisk.ParseContext, requireSize bool) api.St
 		}
 	}
 
+	if (c.repubSource != "" && c.repubDest == "") || (c.repubSource == "" && c.repubDest != "") || (c.repubHeadersOnly && (c.repubSource == "" || c.repubDest == "")) {
+		fisk.Fatalf("must specify both --republish-source and --republish-destination")
```
ripienaar, line 2316: we check headers here but dont mention it in the error so might need to handle the headers only case seperately with a more, specific, descriptive error for if its set

cli/stream_command.go
```diff
@@ -2232,6 +2324,17 @@ func (c *streamCmd) prepareConfig(_ *fisk.ParseContext, requireSize bool) api.St
 		}
 	}
 
+	if (c.subjectTransformSource != "" && c.subjectTransformDest == "") || (c.subjectTransformSource == "" && c.subjectTransformDest != "") {
+		fisk.Fatalf("must specify both --transform-source and --transform-destination")
+	}
+
+	if c.subjectTransformSource != "" && c.subjectTransformDest != "" {
```
ripienaar, line 2331: this if is redundant then - you would always error out above if this case is false.

ripienaar (changes requested):
very very close now :)

cli/stream_command.go
jnmoyne: It's getting late and it's been a long day 😅, actually done.

cli/stream_command.go
jnmoyne: I was wondering about that, added now

cli/stream_command.go
jnmoyne: The test is not there to check both are not "" (that's taken care of in the checkRepubTransform) but to check if they are set or not in which case they need to be included in the generated cfg.

I could check just on one since I know if one is set the other is as well, but that felt 'not right' so I did that check on both 🤷‍♂️

ripienaar: ok yeah you're right, not working for 3 months rots the brain :)

ripienaar (approved):
LGTM

cli/stream_command.go
jnmoyne: If that's the reason, then I'm jealous of your brain rot :)

PR #709, opened 2023-02-20
Removed --no-queue option from /cli/cheats/bench.md because it's not …

davedotdev:
`--no-queue` is no longer a CLI option and therefore removing it from /cli/cheats/bench.md.


ripienaar (approved):
LGTM

PR #727, opened 2023-03-07
Add support for microseconds in humanizeDuration

codegangsta:
Just a minor addition, but ran into it when testing `nats micro` locally. Wasn't able to get anything meaningful when services have sub millisecond response times.


ripienaar (approved):
LGTM

PR #731, opened 2023-03-12
Add support for sub --wait and bind to existing durable

bruth:
This was motivated by wanting to write a script without blocking indefinitely.

Of course, feel free suggest an alternative approach or reject.

### DIFF
--- a/cli/sub_command.go
+++ b/cli/sub_command.go
@@ -122,9 +122,12 @@ func (c *subCmd) subscribe(p *fisk.ParseContext) error {
 	)
 	defer cancel()
 
+	// If the wait timeout is set, then we will cancel after the timer fires.
 	var t *time.Timer
 	if c.wait > 0 {
-		t = time.NewTimer(c.wait)
+		t = time.AfterFunc(c.wait, func() {
+			cancel()
+		})
 		defer t.Stop()
 	}
 
@@ -181,13 +184,9 @@ func (c *subCmd) subscribe(p *fisk.ParseContext) error {
 		}
 
 		// Check if we have timed out.
-		if t != nil {
-			// Attempt to reset the timer. If false, we have timed out.
-			if !t.Reset(c.wait) {
-				sub.Unsubscribe()
-				cancel()
-				return
-			}
+		if t != nil && !t.Reset(c.wait) {
+			cancel()
+			return
 		}
 	}
 
@@ -342,17 +341,7 @@ func (c *subCmd) subscribe(p *fisk.ParseContext) error {
 		return err
 	}
 
-	var tc <-chan time.Time
-	if t != nil {
-		tc = t.C
-	} else {
-		tc = make(chan time.Time)
-	}
-
-	select {
-	case <-ctx.Done():
-	case <-tc:
-	}
+	<-ctx.Done()
 
 	return nil
 }
--- a/cli/sub_command.go
+++ b/cli/sub_command.go
@@ -261,8 +261,19 @@ func (c *subCmd) subscribe(p *fisk.ParseContext) error {
 			opts = append(opts, nats.HeadersOnly())
 		}
 
+		// Check if the durable exists and ignore all the other options.
+		var bindDurable bool
 		if len(c.durable) > 0 {
-			opts = append(opts, nats.Durable(c.durable))
+			con, err := js.ConsumerInfo(c.stream, c.durable)
+			if err == nil {
+				bindDurable = true
+				c.jsAck = con.Config.AckPolicy != nats.AckNonePolicy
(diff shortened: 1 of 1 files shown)


cli/consumer_command.go
```diff
@@ -207,6 +209,8 @@ func configureConsumerCommand(app commandHost) {
 	consSub.Flag("ack", "Acknowledge received message").Default("true").BoolVar(&c.ack)
 	consSub.Flag("raw", "Show only the message").Short('r').UnNegatableBoolVar(&c.raw)
 	consSub.Flag("deliver-group", "Deliver group of the consumer").StringVar(&c.deliveryGroup)
+	consSub.Flag("max-wait", "Max time to wait before unsubscribing.").DurationVar(&c.maxSubWait)
+	consSub.Flag("max-count", "Max number of messages to consume before unsubscribing.").IntVar(&c.maxSubCount)
```
ripienaar, line 213: kind of want to drop `consumer sub`, its a bit gross.  I think `nats sub` with its jetstream features maybe better to focus on

ripienaar: let me know if you can add there or maybe sub doesnt do something that c sub does then fine we add it here for now

bruth: Ah I missed `nats sub` supports JS now. I can move this over there.

bruth: I see that `nats sub` already supports `count` which is nice. The only issue/diff I see is that if a durable already exists, it does not attempt to bind rather than building up the set of options. I will attempt to cleanly handle that case.

bruth: ^And the issue I see, for example.. I created a push durable with ack explicit, but the options use `AckNone`, otherwise the "upsert" behavior could have worked.

ripienaar: sounds good thanks

ripienaar: Yeah, neither nats sub nor consumer sub is good enough at this point :)

bruth:
Alright, moved to sub command. Added `--wait` option and support for binding to an existing durable.

cli/sub_command.go
```diff
@@ -240,8 +261,19 @@ func (c *subCmd) subscribe(p *fisk.ParseContext) error {
...
 
+		// Check if the durable exists and ignore all the other options.
+		var bindDurable bool
 		if len(c.durable) > 0 {
-			opts = append(opts, nats.Durable(c.durable))
+			con, err := js.ConsumerInfo(c.stream, c.durable)
+			if err == nil {
+				bindDurable = true
+				c.jsAck = con.Config.AckPolicy != nats.AckNonePolicy
+				log.Printf("Subscribing to JetStream Stream %q using existing durable %q", c.stream, c.durable)
```
ripienaar, line 271: I think we can simplify this a bit:

rather than this `bindDurable` and the later if around `bindDurable` we can here set `c.subject=""` and just append the `nats.Bind()` call to `opts`.

Rename `bindDurable` to `shouldAck` and later `c.jsAck=shouldAck`.

no extra logic then and just one call to `js.Subscribe()`

cli/sub_command.go
```diff
@@ -306,7 +342,17 @@ func (c *subCmd) subscribe(p *fisk.ParseContext) error {
...
+	if t != nil {
+		tc = t.C
+	} else {
+		tc = make(chan time.Time)
+	}
+
+	select {
+	case <-ctx.Done():
+	case <-tc:
+	}
```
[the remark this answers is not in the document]
bruth: Just to clarify the intent of "wait," it is the wait time _after_ the last message received (maybe timeout is a better option name). So that is why it gets reset any time a message is received.

For the `done` chan, I suppose I could also simply call `cancel()` from within the `AfterFunc` function? It could be reworked that once `<-ctx.Done()` is received on, `sub.Unsubscribe` is called as well.

cli/sub_command.go
bruth: When I tried that before, the additional options seem to cause issues even if `nats.Bind(...)` is set. e.g.

```
$ nats sub --stream bar --durable foo
16:27:37 Subscribing to JetStream Stream "bar" using existing durable "push"
nats: error: configuration requests heartbeat to be 5s, but consumer's value is 1s
```

cli/sub_command.go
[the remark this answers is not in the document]
ripienaar: Right missed that subtlety, this makes more sense now sorry about that.

Yes could just call cancel() indeed, unsubscribe is kind of redundant since we're exiting - would be a wasted round trip.

cli/sub_command.go
ripienaar: ah thats unfortunate :( really dislike this api...leave as is then thanks

ripienaar (approved):
LGTM

Issue #732, opened 2023-03-17
nats server run with ngs context fails on windows


kmpm:
I'm running then natscli compiled from source 0d8f405e1130eb750cb3ad90dba0be982228d969 (because wanting to play with nats.micro) and are having issues following
the guide at https://docs.synadia.com/guides/ngs-dayone-guide#living-on-the-edge-ngs-leaf-nodes

One of the steps would have me to run the embedded server but it fails.
```powershell
nats server run --context ngsleafnode --jetstream --extend
nats: error: Parse error on line 51: 'Invalid escape character 'U'. Only the following escape characters are allowed: \xXX, \t, \n, \r, \", \\.'
```

This looks a lot like issue #486 but my guess it is having a hard time finding the credentials or something because of lots of backslashes that are not escaped correctly.

```powershell
nats context info ngsleafnode
NATS Configuration Context "ngsleafnode"

      Description: synadia (Synadia Communications Inc.)
      Server URLs: tls://connect.ngs.global
      Credentials: C:\Users\<redacted username>\.local\share\nats\nsc\keys\creds\synadia\<redacted account>\default.creds (OK)
             Path: C:\Users\<redacted username>\.config\nats\context\ngsleafnode.json
```

ripienaar:
You should be able to swop the backslashes for forward slashes and it should work, i think

kmpm:
Yes it works if I edit the context. 
So is the fault with `nsc` that created the context or `natscli` where it is used?
The context is technically correct since windows use backslash but where would you like the issue fixed?

ripienaar:
Can you paste your context file please? `C:\Users\<redacted username>\.config\nats\context\ngsleafnode.json`

kmpm:
I have traced the issue to somewhere in the `parse` func in  `nats-server/conf/parse.go`

The context looks like this...
```
# Friendly description for this context shown when listing contexts
description: synadia (Synadia Communications Inc.)

# A comma separated list of NATS Servers to connect to
url: tls://connect.ngs.global

# Connect using a specific username, requires password to be set
user: ""
password: ""

# Connect using a NATS Credentials stored in a file
creds: C:\Users\<redacted username>\.local\share\nats\nsc\keys\creds\synadia\<redacted account>\default.creds

# Connect using a NKey derived from a seedfile
nkey: ""

# Configures a token to pass in the connection
token: ""

# Sets a x509 certificate to use, both cert and key should be set
cert: ""
key: ""

[...]

PR #736, opened 2023-03-17
quote StoreDir and reuse new escape filter

kmpm:
As a follow up to #732 

- Reuse the new escape filter for .StoreDir instead of separate fix.
- Quote store_dir value

### DIFF
--- a/cli/server_run_command.go
+++ b/cli/server_run_command.go
@@ -55,7 +55,7 @@ logtime: false
 
 {{- if .JetStream }}
 jetstream {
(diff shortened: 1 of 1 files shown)


kmpm:
Signed, if SSH signing is enough.

ripienaar:
Should be enough you just need to upload your pubkey to GitHub so it can verify

kmpm:
It is now, and the check on the pr cleared

ripienaar:
Weird, maybe some caching issue at github, anyway will merge as is.  Thanks a lot!

PR #737, opened 2023-03-17
Allow to ignore prompt for stream info

ramonberrutti-f3:
If the stream that we want retrieve information doesn't exist, we prompt with a list of multiple stream. 
Adding the force command that is already implemented but the c.force never be populated.

```go
	if force {
		return "", nil, fmt.Errorf("unknown stream %q", stream)
	}
```

### DIFF
--- a/cli/consumer_command.go
+++ b/cli/consumer_command.go
@@ -163,6 +163,7 @@ func configureConsumerCommand(app commandHost) {
 	consInfo.Arg("stream", "Stream name").StringVar(&c.stream)
 	consInfo.Arg("consumer", "Consumer name").StringVar(&c.consumer)
 	consInfo.Flag("json", "Produce JSON output").Short('j').UnNegatableBoolVar(&c.json)
+	consInfo.Flag("force", "Force info without prompting").Short('f').UnNegatableBoolVar(&c.force)
 
 	consAdd := cons.Command("add", "Creates a new Consumer").Alias("create").Alias("new").Action(c.createAction)
 	consAdd.Arg("stream", "Stream name").StringVar(&c.stream)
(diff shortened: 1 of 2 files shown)


cli/consumer_command.go
```diff
@@ -163,6 +163,7 @@ func configureConsumerCommand(app commandHost) {
 	consInfo.Arg("stream", "Stream name").StringVar(&c.stream)
 	consInfo.Arg("consumer", "Consumer name").StringVar(&c.consumer)
 	consInfo.Flag("json", "Produce JSON output").Short('j').UnNegatableBoolVar(&c.json)
+	consInfo.Flag("force", "Force info without prompting").Short('f').UnNegatableBoolVar(&c.force)
```
ripienaar, line 166: I think force is maybe the wrong flag for this - I have been considering a general higher level flag that would apply equally to streams, consumers and a few other places that disables the lists since we have this annoyance a few places.

So a top level flag for --no-select maybe that then applies to all the various places.  That would need to feed into a new option setting.

I can take care of that if its too much for you to tackle?

ramonberrutti-f3: yes of course. our idea is to use nats cli for some back scripts.

ripienaar: OK will see about that later this week

ripienaar: I think though if thsoe scripts run without a TTY it wont prompt

ripienaar: Lets just change these 2 force flags to:

```
consInfo.Flag("select", "Select consumers from a list").Default("true").UnNegatableBoolVar(&c.force)
```

this will be a bit less confusing than force - but then we just reuse the force boolean, then its good to go I think.

So you'd just pass --no-select

ripienaar:
Done in https://github.com/nats-io/natscli/pull/749

Issue #636, opened 2022-11-18
Supporting gob stream encoder?


oderwat:
I consider using some kind of gob encoding for the payload (see `GOB_ENCODER`) and think that it would be nice to have something like `nats sub --gob gobsubject` which then dumps the result in some more readable way. I also wonder if there should be a `--hex` or `--b64` flag to output the data as hex or base64 encoded data. This probably would help when debugging something which is not using JSON or the default encoder. As these tools are really cool, I wonder it this was considered before and what the state is for that.

ripienaar:
I have considered it - but there is an infinite amount of options so seems like a losing battle

so it supports emitting the data raw (no decorations or metadata) and you should be able to pipe or to other, encoding specific, tools or use things like xargs. 

If anything rather than a long series of supported encoding formats I am more likely to add the ability to pass each message through an external binary for deciding before rendering so it can support anything.

oderwat:
> If anything rather than a long series of supported encoding formats I am more likely to add the ability to pass each message through an external binary for deciding before rendering so it can support anything.

I see, and I also thought about piping it through some decoder (and would implement one right away). 

Still, using b64 / hex seems like an easy way to support everything for a quick "copy & paste" or in the case of hex even for reading it "natively". While "gob" is one of the NATS supported internal decoders (at least I am coming right from an example that shows its usage). I have no insights, but it feels like a practical option for encoding when staying in the Go language domain for clients and services.

> so it supports emitting the data raw (no decorations or metadata)

Furthermore, the current implementation just spills everything into the terminal, even if you do not use `--raw`. This may create a big mess. Having it detect binary and convert that to some hex representation (or suppress it) seems very attractive to me.

ripienaar:
Yeah it’s certainly not ideal at present with binary data (though iirc I do base64 it? Might remeber wrong)

hex and base64 decoders at the least would be good. But I would need to be carefully to detect if the output is not a terminal to enable that behaviour since it will almost certainly be binary data then

oderwat:
I wonder if some people already rely on it being binary in the non `--raw` mode. I think the standard mode could get some variant of the raw mode, maybe `--unfiltered` or `--noesc`. But perhaps it would be safer to add a `--escape` flag, that escapes the output (into base64 or hex) only if it is not printable because if you just leave it "raw" in the case where it is not a terminal, it will not work well with something like `tee` or `watch tail -100` or something I need to make up :)

oderwat:
[...]

PR #739, opened 2023-03-18
Adding the possibility to translate the msg.Data before output.

oderwat:
We have decided to use the nats-cli tool to assist in debugging and monitoring NATS stream messages stored in cbor format. However, we would prefer the messages displayed in an easier-to-read formatting. The same goes for JSON, which we would like to have a better readable representation too.

In response to my inquiry #636 about adding an encoder/decoder, @ripienaar suggested that this could be achieved by passing the output through a CLI utility that performs the translation. This would allow users to add their own translators or formatters as needed.

I have created a proof of concept implementation for this approach, and I hope that it will be incorporated into the tool in the future.

I added a template for `{{subject}}` which can be used to let the translator know about the subject, so one could write it in a way that it uses internal knowledge to assist in decoding the data without the need to add information to the messages. There could also be a template for header fields or maybe also other information (the stream name, for example). I didn't want to add much more until the general idea gets the green light.

### DIFF
--- a/cli/filter_data_through_cmd.go
+++ b/cli/filter_data_through_cmd.go
@@ -1,0 +1,62 @@
+package cli
+
+import (
+	"bytes"
+	"fmt"
+	"github.com/google/shlex"
+	"os/exec"
+	"strings"
+	"text/template"
+)
+
+func outPutMSGBody(data []byte, filter, subject, stream string) {
+	if len(data) == 0 {
+		fmt.Println("nil body")
+		return
+	}
+
+	data, err := filterDataThroughCmd(data, filter, subject, stream)
+	if err != nil {
+		// using q here so raw binary data will be escaped
+		fmt.Printf("%q\nError while translating msg body: %s\n\n", data, err.Error())
+		return
+	}
+	output := string(data)
+	fmt.Println(output)
+	if !strings.HasSuffix(output, "\n") {
+		fmt.Println()
+	}
+}
+
+func filterDataThroughCmd(data []byte, filter, subject, stream string) ([]byte, error) {
+	if filter == "" {
+		return data, nil
+	}
+	funcMap := template.FuncMap{
+		"Subject": func() string { return subject },
+		"Stream":  func() string { return stream },
+	}
+
+	tmpl, err := template.New("translate").Funcs(funcMap).Parse(filter)
+	if err != nil {
+		return nil, err
+	}
+	var builder strings.Builder
+	err = tmpl.Execute(&builder, nil)
+	if err != nil {
+		return nil, err
+	}
+
+	parts, err := shlex.Split(builder.String())
+	if err != nil {
+		return nil, fmt.Errorf("the filter command line could not be parsed: %w", err)
+	}
+	cmd := parts[0]
+	args := parts[1:]
+
+	runner := exec.Command(cmd, args...)
+	// pass the message as string to stdin
+	runner.Stdin = bytes.NewReader(data)
+	// maybe we want to do something on error?
+	return runner.CombinedOutput()
+}
--- a/cli/stream_command.go
+++ b/cli/stream_command.go
@@ -134,6 +134,7 @@ type streamCmd struct {
 	vwStartDelta time.Duration
 	vwPageSize   int
 	vwRaw        bool
+	vwTranslate  string
 	vwSubject    string
 
 	dryRun         bool
@@ -300,6 +301,7 @@ func configureStreamCommand(app commandHost) {
 	strView.Flag("id", "Start at a specific message Sequence").IntVar(&c.vwStartId)
 	strView.Flag("since", "Delivers messages received since a duration like 1d3h5m2s").DurationVar(&c.vwStartDelta)
 	strView.Flag("raw", "Show the raw data received").UnNegatableBoolVar(&c.vwRaw)
+	strView.Flag("translate", "Translate the message data by running it through the given command before output").StringVar(&c.vwTranslate)
 	strView.Flag("subject", "Filter the stream using a subject").StringVar(&c.vwSubject)
 
 	strGet := str.Command("get", "Retrieves a specific message from a Stream").Action(c.getAction)
@@ -307,6 +309,7 @@ func configureStreamCommand(app commandHost) {
 	strGet.Arg("id", "Message Sequence to retrieve").Int64Var(&c.msgID)
 	strGet.Flag("last-for", "Retrieves the message for a specific subject").Short('S').PlaceHolder("SUBJECT").StringVar(&c.filterSubject)
 	strGet.Flag("json", "Produce JSON output").Short('j').UnNegatableBoolVar(&c.json)
+	strGet.Flag("translate", "Translate the message data by running it through the given command before output").StringVar(&c.vwTranslate)
 
 	strBackup := str.Command("backup", "Creates a backup of a Stream over the NATS network").Alias("snapshot").Action(c.backupAction)
 	strBackup.Arg("stream", "Stream to backup").Required().StringVar(&c.stream)
@@ -702,15 +705,7 @@ func (c *streamCmd) viewAction(_ *fisk.ParseContext) error {
 			}
 
 			fmt.Println()
-			if len(msg.Data) == 0 {
-				fmt.Println("nil body")
-			} else {
-				fmt.Println(string(msg.Data))
-				if !strings.HasSuffix(string(msg.Data), "\n") {
-					fmt.Println()
-				}
-			}
-
+			outPutMSGBody(msg.Data, c.vwTranslate, msg.Subject, meta.Stream())
 		}
 
 		if last {
@@ -2705,9 +2700,7 @@ func (c *streamCmd) getAction(_ *fisk.ParseContext) (err error) {
 		}
 		fmt.Println()
 	}
-
-	fmt.Println(string(item.Data))
-	fmt.Println()
+	outPutMSGBody(item.Data, c.vwTranslate, item.Subject, c.stream)
 	return nil
 }
 
--- a/cli/sub_command.go
+++ b/cli/sub_command.go
@@ -38,6 +38,7 @@ type subCmd struct {
 	queue                 string
 	durable               string
 	raw                   bool
+	translate             string
 	jsAck                 bool
 	inbox                 bool
 	match                 bool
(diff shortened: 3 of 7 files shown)


cli/filter_data_through_cmd.go
```diff
@@ -0,0 +1,93 @@
...
+		args[idx] = strings.ReplaceAll(arg, "{{subject}}", subject)
+	}
+	runner := exec.Command(cmd, args...)
+	// pass the message as string to stdin
+	runner.Stdin = bytes.NewReader(data)
+	// maybe we want to do something on error?
+	return runner.CombinedOutput()
+}
+
+func parseCommandLine(command string) ([]string, error) {
```
ripienaar, line 29: maybe https://github.com/google/shlex can be used, despite being archived its still useful think it does what you're doing here

ripienaar (commented):
This is pretty cool, good use case examples also.

I left one comment there let me know what you think otherwise I will take a closer look next week

cli/filter_data_through_cmd.go
oderwat: I was not sure about adding dependencies. But you are right, it will be much better to use this as it has a test and is doing exactly what it is used for here.

oderwat:
I added a 'stream' template and pondered about adding the headers, as one could use them especially to describe the data. 

I am also unsure of the argument name "translate", in our code base this would probably something like "cvt" or "flt".


**The following is not part of this PR**

I noticed that the code has many missing error checks ("will never happen" / "if that breaks, I am fleeing the country already"). We made a simple "must" package (using generics) to create `must.Ok()`, `must.OkOne()`, `must.OkSkipOne()` (and so on) which simply check and panic or return/ignore one return value. 

We also have an alternative package that checks for the errors and on error lists just the file and line of the caller in the source and `os.Exit(1)`. This is what we use for CLI tooling that is used by people that freak out when they get a panic with stack-dump. :)

This way it is effortless and readable to cover all error checks without having all the single if statements for the error checks. I would suggest adding something like this and could volunteer.

ripienaar:
Hmm mind pointing at a few missing error checks? 

We should probsbly add templates for all the jetstream metadata, not sure it becomes quite hard to use - my first thought was just to dumb the message as json but your examples are quite compelling and json woukd CI okicsge matters in your examples.

oderwat:
> Hmm mind pointing at a few missing error checks?

There are quite a lot

cbvb-599D5214-1CF1-40B6-922D-DE95D115F624

ripienaar:
That’s a test :) fairly acceptable imo as following assertions would not pass had there been an error. 

There are in general vast amounts of unchecked errors you are right - like fmt.Printf returns unchecked errors. So I would be curious if one’s if actual consequence outside of tests.

oderwat:
As usual, adding something like that opens up plenty of ideas. I would rather go with a minimalistic approach first and see what emerges from that. It should be easy to add more later on. For adding something like JSON information, one may create a temp file with that JSON, add the filename as template and remove the JSON after the translator has run.

oderwat:
> That’s a test :) fairly acceptable imo as following assertions would not pass had there been an error.

I would add them, especially in tests. That makes it easier to see where it breaks if it breaks.

This is what I get when I quickly filter out '*_test.go':

cbvb-5D7A9751-689A-4718-80C9-A914F079DC6B

This only comes to my mind because we have a zero warning policy for all code (using GoLand) and I also plan a global golangci-lint config to be mandatory for all pushed code into the CI/CD chains. But I know that I am quite strict with that in our company. We should have more tests though :)

ripienaar:
In this case I would fetch that command. It’s old old code that was just copy and pasted in here.

I wouldn’t want to cha get the code style at all. Rather just add checks where they are of consequence.

oderwat:
I really didn't want to steer up the error handling stuff. As mentioned earlier. This PR was made to extend the ways of using the wonderful nats-cli tool with data that's encoded in some binary format. I actually wrote some "universal translator" implementation today that also uses AVRO model info stored in the stream to decode messages. It helps already by allowing to use "hex", "raw escaped strings", "CBOR", "AVRO" and "JSON" formats. Using "GOB" turned out to neither be smaller than JSON nor is debug printable without recompiling the tools and adding all the structures (or writing a debugging decoder from scratch). I didn't use AVRO before, but found that this may be the best way for us to store large amounts of data consisting of many messages (which makes gob rather bloated). It seems that Kafka has a point in using it. The possibility to store the structure as text and don't need to recompile when updating the format, makes it quite interesting when storing many messages with the same (binary) format.

oderwat:
About that "must" package:

I like how the code looks when using it while prototyping or even for production:

```go
dec := cbor.NewDecoder(os.Stdin)
var cborNative any
must.Ok(dec.Decode(&cborNative))
mappedKeys := must.OkOne(convertMapKeysToStrings(cborNative))
jsonData := must.OkOne(json.MarshalIndent(mappedKeys, "", "  "))
fmt.Printf("%s\n", jsonData)
```

We have now two packages. One that just uses panic() for the errors and another one that uses log/slog and `os.Exit()`. 
Because of this conversation, I decided to make them both exchangeable. I find that much better than having the `if(err) { panic(err) }`. We do forbid variable shadowing too, and as a result, you need to name many of your errors when using just "err" all the time.

ripienaar:
I'll review the data translation towards end of week, thank you.

I would not be interested in anything like the `must` package here, happy to have PRs adding idiomatic go error handling where its lacking now.

cli/filter_data_through_cmd.go
```diff
@@ -0,0 +1,29 @@
...
+	if err != nil {
+		return nil, fmt.Errorf("the filter command line could not be parsed: %w", err)
+	}
+	cmd := parts[0]
+	args := parts[1:]
+	// maybe we want to replace some strings to give the filter something to reason about?
+	for idx := range args {
+		arg := &args[idx]
+		*arg = strings.ReplaceAll(*arg, "{{subject}}", subject)
+		*arg = strings.ReplaceAll(*arg, "{{stream}}", stream)
```
ripienaar, line 22: elsewhere we use `text/template` for this and people would already be used to being able to put spaces in and so forth.  So lets also use `text/template` here and then make it `Subject` and `Stream` also for consistency with like whats in use in `nats req` and so forth wrt capitalization

cli/stream_command.go
```diff
@@ -704,9 +706,18 @@ func (c *streamCmd) viewAction(_ *fisk.ParseContext) error {
 			if len(msg.Data) == 0 {
 				fmt.Println("nil body")
 			} else {
-				fmt.Println(string(msg.Data))
-				if !strings.HasSuffix(string(msg.Data), "\n") {
-					fmt.Println()
+				data := msg.Data
+				var errFilter error
```
ripienaar, line 710: no real value in making new error variables here, just use the one we have. The logic here seems duplicated 3 times I think we can probably extract it into a generic helper function? Also worth imprving on the errors that wold jsut be `Error: ...` a better message would be clearer

ripienaar (changes requested):

cli/filter_data_through_cmd.go
ripienaar: Using text/template also means we can easily extend later

cli/stream_command.go
oderwat: We don't allow variable shadowing in our codebase. Having names for errors also makes the code more readable, so this is out of habit. I can change that of course.

cli/filter_data_through_cmd.go
oderwat: I need to read up in text/template. I never used that myself.

oderwat: I changed it to use the `text/template` package. Please check. I wonder about how there is sometimes `{{.Count}}` (README.me:145) but mostly `{{Count}}` in the code. Am I supposed to use the functions instead of the fields or doing both, what I think that is being done in `pubReplyBodyTemplate()`?

ripienaar: I think the readme is probably wrong. Can pass in a map[string]any wirh those fields is enough and then highlighting them in help output.

oderwat: Do you mean additionally to the `{{.Subject}}` or instead? I ask because `pubReplyBodyTemplate()` allows both types.

ripienaar: you can pick one, which ever you like - the pub/reply one is complex because of a early backward incompatible change and so i made it kind of work.  since yours is new you can just go with one as long as its in --help

oderwat: Well. I picked the `{{.Subject}}` type and updated the README with the last commit already. The reason was that using `{{.Subject}}` instead of `{{Subject}}` may people aware of that it is using the text/template and let them use other functionality of it. But I got unsure because most of the documentation does not use this variant and the additional `.` may make people wonder. So, the `{{Subject}}` variant it will be then. The documentation was updated.

ripienaar: looks good thanks, any chance we can get that duplicated code pulled into a seperate function?

oderwat: I am unsure what duplicate code you refer to.

cli/sub_command.go
```diff
@@ -456,8 +458,18 @@ func prettyPrintMsg(msg *nats.Msg, headersOnly bool) {
 	}
 
 	if !headersOnly {
-		fmt.Println(string(msg.Data))
-		if !strings.HasSuffix(string(msg.Data), "\n") {
+		data := msg.Data
+		if filter != "" {
+			var err error
+			data, err = filterDataThroughCmd(msg.Data, filter, msg.Subject, "")
```
ripienaar, line 464: this code is essentially the same just used 3 times, it can probably be made a function.

And improve the error to be more descriptive

oderwat: Well, I try to avoid calling functions when not needed. Of course, one could call it independent of what filter contains and just return data unmodified if no filter is given. I didn't like that, though. Is that what you want?

In my experiments with it, I found the errors that show up satisfying. What is on your mind? Adding something about the error being caused by the translation using a wrapper?

ripienaar: I mean this whole section of code the with error handling etc, its all duplicated more or less equally in 3 places.  see after lines 710, 2720 and here, its all very much the same - especially the first two.  I am just curious if it can be extracted since later if someone has to introduce a change in how this work its unlikely they will spot the 3 places thats similar, one might be forgotten and then introduces a bug.

"Error: x" isnt clear, error while doing what?

oderwat: I extended the error to tell where it happens, besides that, there are some minor differences that make it difficult to reduce that code. I would let it as it is right now. Maybe doing some refactoring about outputting the messages in general could be made in another step?

BTW: It just realized that running the template on the individual arguments is not needed, and when done before the `shlex` step, you can even use the template language to run different commands based on the data. I added a new commit that

oderwat: Well. I looked at the code for the output again and see what you mean with the duplication. I guess that can be handled. Adding the "\n" can be done every time. I dislike printing the error from inside the "helper" function. But one could refine that as the `msgOutput()` function and handle the error inside that.

oderwat:
OK. I consolidated the code and think that the generalized output function should result in the same user experience while unifying the edge cases (like nil message body). I did not rename the source file which contains that code to make it easier to see my changes. I guess it should be renamed to "output_msg_body.go" before merging.

What do you think about it overall now?

ripienaar (approved):

ripienaar:
Good stuff thank you, do you mind signing this with gpg or your ssh key? Please squash to one commit and sign that commit.

oderwat:
> Good stuff thank you, do you mind signing this with gpg or your ssh key? Please squash to one commit and sign that commit.

I actually never used signing before. I hope I did that right.

ripienaar:
Yup signature looks good

ripienaar:
Thank you

PR #741, opened 2023-03-20
Issues with following Match requests and replies section

mcp5:
I updated line 194 so that it can be copied to the terminal.

for the output of the first command line - nats sub --match-replies cli.demo, I am getting the below, so not the same as stated.  And I am having trouble following the rest of this section. (base) marthapena@Marthas-MacBook-Pro ~ % nats sub --match-replies cli.demo 10:15:57 Matching replies with inbox prefix _INBOX.> 10:15:57 Subscribing on cli.demo

### DIFF
--- a/README.md
+++ b/README.md
@@ -191,7 +191,7 @@ hello headers
 ### match requests and replies
 We can print matching replay-requests together
 ```
-sub --match-replies subject.name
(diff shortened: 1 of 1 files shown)


ripienaar:
thank you

PR #744, opened 2023-03-23
Propagate timeout from the CLI flags to the JetStream client

jperez-igz:
## Description
Propagate the timeout command line flag to wait time value of JetStream client.

## Motivation and Context
We need this change in order to be able to perform the delete from ObjectStore because now this wait time is set by default to 5s and we are not able to modify it from the command line.

## Types of changes
- [X] Bug fix (non-breaking change which fixes an issue)


ripienaar:
Thank you

PR #760, opened 2023-04-15
I forgot to add the translate feature to the req command.

oderwat:
While continuing our [AvroX](https://github.com/metatexx/avrox) development, we tripped over the missing --translation for `nats req` so I added it.


ripienaar:
Thank you

PR #767, opened 2023-04-28
add revision to kv get trace output

maxarndt:
This PR adds the sequence number of a message to the output of `nats kv get <bucket> <key> --trace`

The timestamp of the message was already present but I missed the sequence number. 
I'm aware that the history command shows this information but for me this is the right place as well for this information.

### DIFF
--- a/cli/kv_command.go
+++ b/cli/kv_command.go
@@ -492,7 +492,7 @@ func (c *kvCommand) getAction(_ *fisk.ParseContext) error {
 		return nil
 	}
 
-	fmt.Printf("%s > %s created @ %s\n", res.Bucket(), res.Key(), res.Created().Format(time.RFC822))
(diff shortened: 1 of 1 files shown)


cli/kv_command.go
```diff
@@ -492,7 +492,7 @@ func (c *kvCommand) getAction(_ *fisk.ParseContext) error {
 		return nil
 	}
 
-	fmt.Printf("%s > %s created @ %s\n", res.Bucket(), res.Key(), res.Created().Format(time.RFC822))
+	fmt.Printf("%s > %s seq %d created @ %s\n", res.Bucket(), res.Key(), res.Revision(), res.Created().Format(time.RFC822))
```
ripienaar, line 495: lets call it `revision: %d` instead, revision is the word people using the APIs would recognise.

ripienaar:
Thank you

PR #773, opened 2023-05-05
Add a connection flush before interrupting

ColinSullivan1:
I noticed that sometimes the nats CLI process would exit before the reply command response was sent; this resolves that issue.

### DIFF
--- a/cli/reply_command.go
+++ b/cli/reply_command.go
@@ -197,6 +197,7 @@ func (c *replyCmd) reply(_ *fisk.ParseContext) error {
 		i++
 
 		if c.limit != 0 && uint(i) == c.limit {
+			nc.Flush()
 			ic <- os.Interrupt
 		}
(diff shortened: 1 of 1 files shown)


ripienaar:
Nice catch, mind signing the commit?

ColinSullivan1:
Hmm... something is not right... I'll try to fix it.

image

ColinSullivan1:
I tried both signatures... 

image

Not sure where to go from here except try another branch/PR.  Or maybe you can just trust that I'm Colin?  :)

ripienaar:
Did you upload the public keys to your GH profile? 

If you did then sure will merge

ColinSullivan1:
Yeah, I added a GPG, will make another SSH sometime this weekend...

ripienaar:
Thanks for trying :)

PR #793, opened 2023-05-23
Show stream compression only if compressed, regardless of store type

neilalexander:


ripienaar (approved):
LGTM

PR #795, opened 2023-06-03
Remove dot at end of "Settable using" help.

sebbbastien:
Most of  stream help messages ending with "Settable using" does not have dot (.) after the option to use.

Some have, this tailing dot is not practical when copy and paste.


ripienaar:
Thank you!

PR #799, opened 2023-06-14
Small improvements to bench command

jnmoyne:
change 2 defaults (storage to file, as that's faster for KV) and multisubjectmax to 100k) and adds jitter when pub sleep is used and more than one publisher to avoid all of the publishers publishing at exactly the same time which is not realistic or desirable.


ripienaar (approved):
LGTM

PR #800, opened 2023-06-14
Fix panic in stream view upon reaching end of data

jnmoyne:
Fix panic on nil pointer (msg = nil) upon reaching the end of data introduced by PR #797

### DIFF
--- a/cli/stream_command.go
+++ b/cli/stream_command.go
@@ -672,8 +672,6 @@ func (c *streamCmd) viewAction(_ *fisk.ParseContext) error {
 		}
 	}()
 
-	shouldTerminate := false
-
 	for {
 		msg, last, err := pgr.NextMsg(ctx)
 		if err != nil {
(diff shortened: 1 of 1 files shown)


cli/stream_command.go
```diff
@@ -672,16 +672,15 @@ func (c *streamCmd) viewAction(_ *fisk.ParseContext) error {
...
 	for {
 		msg, last, err := pgr.NextMsg(ctx)
 		if err != nil {
 			if !last {
 				return err
 			}
 			// later we know we reached final last after showing the final message
-			shouldTerminate = true
+			log.Println("Reached apparent end of data")
+			return nil
```
ripienaar, line 683: This is how it was before but that’s actually a bug - note the last message is never rendered we bail before rendering in cases where pending ==. 0

ripienaar:
Included the fix here https://github.com/nats-io/natscli/pull/804, will investigate the other thing

PR #828, opened 2023-07-23
Merge identical switch cases

mdawar:
Just a small change, merged identical switch cases checking the operation type in the watch action.


mdawar:
I have signed the commit and force pushed.

ripienaar:
Thank you!

PR #829, opened 2023-07-25
server report js: add an "Err %" column

philpennock:
When one server is much busier, the "API Err" cell might look uncommonly high,
but the actual error rate is not higher.  Having an "Err %" column should make
that much clearer.

### DIFF
--- a/cli/server_report_command.go
+++ b/cli/server_report_command.go
@@ -268,6 +268,11 @@ func (c *SrvReportCmd) reportJetStream(_ *fisk.ParseContext) error {
 		if renderDomain {
 			row = append(row, js.Data.Config.Domain)
 		}
+		errCol := f(jss.API.Errors)
+		if jss.API.Total > 0 && jss.API.Errors > 0 {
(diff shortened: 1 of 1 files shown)


ripienaar:
Rather than a column that takes a lot of space let’s add just “123 / 5%” in the existing g one. 

Tbh people shouldn’t read much/anything into error rates. Even a stream info for a stream that don’t exist results in an error. And in a cluster every stream info is sent to all servers so rates go up for streams that aren’t everywhere. It’s a very misleading number as it stands

philpennock:
Yeah, it's not useful, but if one server is seeing much more traffic and so the error count is high, we had people pointing to the high error number and wondering what was wrong with that one server.  Having the rate makes it clearer when there's nothing special about a particular server and avoids people spending time on the wrong thing.

In the update, I've moved the rate in, and I only add it if the error is non-zero, to reduce clutter a little.

philpennock:
oops, missed that this repo now requires signed commits.  Rebase/squashed to one commit and is signed, same code as the previous two commits.

Issue #834, opened 2023-07-27
Context saved with a token causes an `Authorization Violation` error


mdawar:
## Defect

Saving a context using `nats` CLI with a token causes an `Authorization Violation` error when using this context.

#### Versions Used
- nats CLI: `v0.0.35`
- nats-server: `v2.9.20`

#### Steps or code to reproduce the issue:
1. Run a NATS server with token authentication:
```sh
docker run --rm -it -p 4222:4222 nats --auth secret
```

2. Save a context with a token:
```sh
nats context save tokenauth --user secret
```

3. Select the saved context:
```sh
nats context select tokenauth
```

4. Run any command
```sh
nats account info
```

You will get the error:
```sh
nats: error: setup failed: nats: Authorization Violation
```

#### Expected result:
The authentication should work just like when using a user/password combination.

#### Actual result:
The authentication fails, because the context is saved with a `user` field too.

```
{
  "description": "",
  "url": "nats://127.0.0.1:4222",
  "token": "secret",
  "user": "secret",
  "password": "",
  "creds": "",
  "nkey": "",
  "cert": "",
  "key": "",
  "ca": "",
  "nsc": "",
  "jetstream_domain": "",
  "jetstream_api_prefix": "",
  "jetstream_event_prefix": "",
  "inbox_prefix": "",
  "user_jwt": ""
}
```

I get this output from the NATS server running in `debug` and `trace` modes:

```sh
- <<- [CONNECT {"verbose":false,"pedantic":false,"user":"secret","tls_required":false,"name":"NATS CLI Version development","lang":"go","version":"1.19.0","protocol":1,"echo":true,"headers":true,"no_responders":true}]
- authentication error
- ->> [-ERR Authorization Violation]
- Client connection closed: Authentication Failure
```

If you set the `user` field to an empty string, the authentication works:
```sh
<<- [CONNECT {"verbose":false,"pedantic":false,"auth_token":"secret","tls_required":false,"name":"NATS CLI Version development","lang":"go","version":"1.28.0","protocol":1,"echo":true,"headers":true,"no_responders":true}]
```

Since the token is passed using the `--user` flag, the context is saved in [`createCommand`](https://github.com/nats-io/natscli/blob/1ba4c811205062e86c16b721979ff3af44aace4d/cli/context_command.go#L502) with the `user` and `token` fields both set to the same value.

I don't know if the solution would be using a new flag for the token or just checking the values before saving.

PR #835, opened 2023-07-27
Do not save the user for a context with token auth

mdawar:
Possible fix for #834 if the token won't have it's own separate flag.


ripienaar:
Thank you!

PR #845, opened 2023-08-19
Implements multi-filter sources and mirrors

jnmoyne:
Implements multi-filter sources and mirrors in stream_command.go

(needs nats-server with this PR https://github.com/nats-io/nats-server/pull/4403 to display info properly)

### DIFF
--- a/cli/stream_command.go
+++ b/cli/stream_command.go
@@ -758,6 +758,10 @@ func (c *streamCmd) viewAction(_ *fisk.ParseContext) error {
 				fmt.Println()
 				for k, vs := range msg.Header {
 					for _, v := range vs {
+						if k == "Nats-Stream-Source" {
+							v = strings.ReplaceAll(v, "\f", "\u21B5")
+						}
+
 						fmt.Printf("  %s: %s\n", k, v)
 					}
 				}
@@ -1863,12 +1867,36 @@ func (c *streamCmd) showStreamInfo(info *api.StreamInfo) {
 
 		if s.SubjectTransformDest != "" {
 			filter := ">"
+
 			if s.FilterSubject != "" {
 				filter = s.FilterSubject
 			}
-			cols.AddRowf("Subject Transform", "%s to %s", filter, s.SubjectTransformDest)
+
+			cols.AddRowf("Subject Filter and Transform", "%s to %s", filter, s.SubjectTransformDest)
 		} else {
-			cols.AddRowIfNotEmpty("Subject Filter", s.FilterSubject)
+			if s.SubjectTransformDest == "" && len(s.SubjectTransforms) == 0 {
+				cols.AddRowIfNotEmpty("Subject Filter", s.FilterSubject)
+			} else {
+				for i := range s.SubjectTransforms {
+					var t string
+
+					if i == 0 {
+						if len(s.SubjectTransforms) > 1 {
+							t = "Subject Filters and Transforms"
+						} else {
+							t = "Subject Filter and Transform"
+						}
+					} else {
+						t = ""
+					}
+
+					if s.SubjectTransforms[i].Destination == "" {
+						cols.AddRowf(t, "%s to [no transform]", s.SubjectTransforms[i].Source)
+					} else {
+						cols.AddRowf(t, "%s to %s", s.SubjectTransforms[i].Source, s.SubjectTransforms[i].Destination)
+					}
+				}
+			}
 		}
 
 		cols.AddRow("Lag", s.Lag)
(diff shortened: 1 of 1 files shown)


cli/stream_command.go
```diff
@@ -1863,12 +1867,36 @@ func (c *streamCmd) showStreamInfo(info *api.StreamInfo) {
...
+
 			if s.FilterSubject != "" {
 				filter = s.FilterSubject
 			}
-			cols.AddRowf("Subject Transform", "%s to %s", filter, s.SubjectTransformDest)
+
+			cols.AddRowf("Subject Filter and Transform", "%s to %s", filter, s.SubjectTransformDest)
 		} else {
-			cols.AddRowIfNotEmpty("Subject Filter", s.FilterSubject)
+			if s.SubjectTransformDest == "" && len(s.SubjectTransforms) == 0 {
```
ripienaar, line 1877: nested ifs break my brain completely, can we make this a single case statement?

cli/stream_command.go
```diff
@@ -2370,12 +2398,56 @@ func (c *streamCmd) askMirror() *api.StreamSource {
...
+				Message: "Subject transform destination",
+				Help:    "Transform the subjects using this destination (hit enter for no transformation)",
+			}, &destination)
+			fisk.FatalIfError(err, "could not request transform destination")
+
+			sources = append(sources, source)
+			destinations = append(destinations, destination)
+		}
+
+		if len(sources) > 0 {
```
ripienaar, line 2434: also make this a case 

```
switch {
case len(sources)==1:
case len(sources)>0:
}
```

cli/stream_command.go
```diff
@@ -2443,18 +2515,53 @@ func (c *streamCmd) askSource(name string, prefix string) *api.StreamSource {
...
```
ripienaar, line 2548: case here too

cli/stream_command.go
```diff
@@ -1863,12 +1867,36 @@ func (c *streamCmd) showStreamInfo(info *api.StreamInfo) {
...
+				for i := range s.SubjectTransforms {
+					var t string
+
+					if i == 0 {
+						if len(s.SubjectTransforms) > 1 {
+							t = "Subject Filters and Transforms"
+						} else {
+							t = "Subject Filter and Transform"
+						}
+					} else {
```
ripienaar, line 1889: else clause not needed `var t string` makes an empty string, this is why I prefer `t := ""` instead when making this kind of var because it becomes much more obvious then

ripienaar (changes requested):
some small changes, thanks!

cli/stream_command.go
jnmoyne: Changed to a switch

cli/stream_command.go
jnmoyne: Done

cli/stream_command.go
jnmoyne: Done

cli/stream_command.go
jnmoyne: Done

cli/stream_command.go
```diff
@@ -1861,14 +1851,38 @@ func (c *streamCmd) showStreamInfo(info *api.StreamInfo) {
...
+					}
+
+					if s.SubjectTransforms[i].Destination == "" {
+						cols.AddRowf(t, "%s to [no transform]", s.SubjectTransforms[i].Source)
+					} else {
+						cols.AddRowf(t, "%s to %s", s.SubjectTransforms[i].Source, s.SubjectTransforms[i].Destination)
+					}
+				}
+			}
+		default:
```
ripienaar, line 1878: purpose of using switch is to remove nested conditionals they evaluate top-down and so great for removing this kind of nested logic

```
switch {
case s.SubjectTransformDest != "": 
  // outer switch default clause

case len(s.SubjectTransforms) == 0:
  // inner switch case 0 clause

case len(s.SubjectTransforms) > 0:
 // inner switch default clause

}```

ripienaar (changes requested):
small change

cli/stream_command.go
jnmoyne: I was actually trying to represent the nested aspect of the logic (i.e. is SubjectTransformDest == "" then look at SubjectTransforms and depending on it's length do this or that) because that's how I represent what I want to do in my head, hence the if/then logic but fair enough maybe that makes it harder for someone else who doesn't know exactly what I want the code to do to follow.

Flattened it to a single switch which is a few lines shorter anyways (and I always like less lines of code)

ripienaar: I definitely take clarity over lines of code :) but yeah it’s just a thing with idiomatic go to try and not nest logic so it always gets me

ripienaar (approved):
LGTM

PR #853, opened 2023-08-29
use global JS settings in bench command

caleblloyd:
Use the global JS settings in calls to `nc.JetStream()` inside of the bench comand


jnmoyne (approved):
LGTM

PR #861, opened 2023-09-19
Attempt to remove the subject_transform_dest struct key

ripienaar:


cli/stream_command.go
```diff
@@ -1204,10 +1204,14 @@ func (c *streamCmd) reportAction(_ *fisk.ParseContext) error {
...
+				if source.FilterSubject == "" {
+					continue
+				}
+
+				if len(source.SubjectTransforms) == 0 {
 					edge.Label(source.FilterSubject)
-				} else if source.FilterSubject != "" && source.SubjectTransformDest != "" {
-					edge.Label(source.FilterSubject + " to " + source.SubjectTransformDest)
+				} else if len(source.SubjectTransforms) == 1 {
+					edge.Label(source.SubjectTransforms[0].Source + " to " + source.SubjectTransforms[0].Destination)
```
ripienaar, line 1214: Seems we never did anything here if SubjectTransforms were set at all? So I limited the change to just if its exactly 1

jnmoyne: I think not handling SubjectTransforms here was simply an omission, so no need to limit it to just == 1, it should just be if > 1

ripienaar: not convinced that would work, because here the label pairs are to->dest and it needs to handle many.  Just using a >1 here would not look good.

I think I will leave it as is for release, we can fix later to handle multiple better

jnmoyne: Maybe put something like "multiple" in the filter and destinations columns if there's more than 1 in SubjectTransforms?

jnmoyne: Then find a way to display more than one in the table format later

ripienaar: These edges are here are used to produce a dot graph, not for the table, its not something to fix today I am afraid.

jnmoyne: 👍

jnmoyne (approved):
LGTM

jnmoyne (approved):
LGTM

PR #877, opened 2023-09-26
Report observer status in cluster replicas for streams and consumers

neilalexander:
This is to support nats-io/nats-server#4582.


go.mod
```diff
@@ -2,9 +2,14 @@ module github.com/nats-io/natscli
 
 go 1.20
 
+replace github.com/nats-io/jsm.go => github.com/nats-io/jsm.go v0.1.1-0.20230926090317-6fb84b8ef01d
```
ripienaar, line 5: I'd rather not have these and instead update the dependency to point to the commit.

I keep forgetting about these and then do a release with them in which then end up failing `go install` installation method - happened when I did 0.1.0 and then had to do 0.1.1.  

So hoping to avoid this mistake all together if I can :)

ripienaar (changes requested):
comment about replace

go.mod
neilalexander: Yep, these will be removed as commented above but for now it's just so that CI does the right thing while in draft state.

ripienaar: Oh, sorry, I missed the comment you left there :)

Issue #879, opened 2023-09-27
Unable to add subjects to existing consumers


jarretlavallee:
### What version were you using?

NATS Server v2.10.1
`natscli` v0.1.0

### What environment was the server running in?

NGS + MacOS

### Is this defect reproducible?

1. Create a consumer with a single filter subject.
2. Edit the consumer to add another subject.
3. Observe the failure. 

```
nats consumer create multi abc --filter a.1 
nats consumer edit multi abc --filter a.1 --filter b.1
```

### Given the capability you are leveraging, describe your expectation?

It should be possible to edit an existing consumer to add another filter subject.

### Given the expectation, what is the defect you are observing?

When trying to edit an existing consumer to add a filter subject, it fails with an error. The error indicates that the wrong API is used. Trace output is below.

```
~ nats consumer edit multi abc --filter a.1 --filter b.1 --trace
08:07:08 >>> $JS.API.CONSUMER.INFO.multi.abc


08:07:08 <<< $JS.API.CONSUMER.INFO.multi.abc
{"type":"io.nats.jetstream.api.v1.consumer_info_response","stream_name":"multi","name":"abc","created":"2023-09-27T14:07:06.233383581Z","config":{"durable_name":"abc","name":"abc","deliver_policy":"all","ack_policy":"explicit","ack_wait":30000000000,"max_deliver":-1,"filter_subject":"a.1","replay_policy":"instant","max_waiting":512,"max_ack_pending":1000,"max_batch":4096,"num_replicas":0},"delivered":{"consumer_seq":0,"stream_seq":0},"ack_floor":{"consumer_seq":0,"stream_seq":0},"num_ack_pending":0,"num_redelivered":0,"num_waiting":0,"num_pending":0,"cluster":{"name":"ngsprod-aws-uswest2","leader":"aws-uswest2-natscj1-1"},"ts":"2023-09-27T14:07:08.592777775Z"}

Differences (-old +new):
  api.ConsumerConfig{
  	... // 7 identical fields
  	Name:           "abc",
  	FilterSubject:  "a.1",
- 	FilterSubjects: []string(Inverse(Sort, []string(nil))),
+ 	FilterSubjects: []string(Inverse(Sort, []string{"a.1", "b.1"})),
  	FlowControl:    false,
  	Heartbeat:      s"0s",
  	... // 18 identical fields
  }
[context] ? Really edit Consumer multi > abc Yes
[...]

PR #881, opened 2023-09-27
[FIXED] consumer edit going from 1 subject filter to more than one

jnmoyne:
Fixes consumer edit going from 1 subject filter to more than one that would move the filters to the `FilterSubjects` array, but not remove the initial single filter from FilterSubject.


jnmoyne:
Fixes: https://github.com/nats-io/natscli/issues/879

ripienaar (approved):
LGTM

PR #883, opened 2023-09-27
[ADDED] `consumer add` when prompting for subject filter(s) can parse the user input as a comma or space separated list of subject filters

jnmoyne:
Added to `consumer add`: when prompting for subject filter(s) can parse the user input as a comma or space separated list of subject filters.


ripienaar:
This is probably not `[ADDED]` but rather whatever we use to indicate fixed.  Personally I dont really want those prefixes in this repo they dont add much value

jnmoyne:
Ok, just in the habit of using those prefixes because that's how they want it for the client libraries, but won't use them for this repo from now on.

PR #886, opened 2023-09-28
Backports a fix for prometheus exit codes on 0.0.35 cli release

ripienaar:


ripienaar:
Merging as tests cant pass with old code

PR #927, opened 2023-11-27
Account management basics

ripienaar:
This adds basic of account management including scoped
signing keys, several todo items left but good for early
feedback


ripienaar:
blocked on https://github.com/synadia-io/jwt-auth-builder.go/pull/24 and update to remove replace directive

PR #934, opened 2023-12-13
feat: skip tls when making connection

amagow-ct:
Add option to skip tls when making nats connection

### DIFF
--- a/cli/util.go
+++ b/cli/util.go
@@ -17,6 +17,7 @@ import (
 	"bufio"
 	"bytes"
 	"context"
+	"crypto/tls"
 	"encoding/base64"
 	"encoding/json"
 	"errors"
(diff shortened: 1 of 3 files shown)


cli/util.go
```diff
@@ -354,6 +355,10 @@ func natsOpts() []nats.Option {
 		connectionName = "NATS CLI Version " + Version
 	}
 
+	if opts.SkipTls {
+		copts = append(copts, nats.Secure(&tls.Config{InsecureSkipVerify: true}))
```
ripienaar, line 359: Would this not break specifying certificates ca etc? or are those handled by nats.go and added to this tlsc?

ripienaar:
Will close this for now as it's old and out dated after no response

PR #951, opened 2024-01-10
fix: changing context name in one command

LautaroJayat:
Changing context name in example, since first is named "local" but then is referenced as "localhost". This might benefit new users of this CLI.


ripienaar:
Thank you :)

Issue #962, opened 2024-01-12
Incorrect local JetStream store directory when $XDG_DATA_HOME is defined


erhhung:
### Observed behavior

When running `nats server run --jetstream`, the **Store Directory** shown in the **JetStream** banner output repeats the "`.local/share/`" path components:

```
[76897] [INF] Starting nats-server
[76897] [INF]   Version:  2.10.0
[76897] [INF]   Git:      [not set]
[76897] [INF]   Name:     nats_development
[76897] [INF]   Node:     HhEiCaYr
[76897] [INF]   ID:       NAIC2EHHLBZSS6H7KZWZRP5U6UVHDB4FNORZT2S733PJMLLGXNIHN6KW
[76897] [INF] Using configuration file: /var/folders/yt/lwq8py0n4qxbkr2mg_nwg1y00000gn/T/nats-server-run-3707093925.cfg
[76897] [INF] Starting JetStream
[76897] [INF]     _ ___ _____ ___ _____ ___ ___   _   __  __
[76897] [INF]  _ | | __|_   _/ __|_   _| _ \ __| /_\ |  \/  |
[76897] [INF] | || | _|  | | \__ \ | | |   / _| / _ \| |\/| |
[76897] [INF]  \__/|___| |_| |___/ |_| |_|_\___/_/ \_\_|  |_|
[76897] [INF]
[76897] [INF]          https://docs.nats.io/jetstream
[76897] [INF]
[76897] [INF] ---------------- JETSTREAM ----------------
[76897] [INF]   Max Memory:      48.00 GB
[76897] [INF]   Max Storage:     1.20 TB
[76897] [INF]   Store Directory: "/Users/erhhung/.local/share/.local/share/nats/nats_development/jetstream"
[76897] [INF] -------------------------------------------
[76897] [INF] Listening for client connections on 0.0.0.0:58137
[76897] [INF] Server is ready
```

Apparently, if the **`XDG_DATA_HOME` environment variable** isn't set, the default store directory is correct:

```sh
$ echo $XDG_DATA_HOME
/Users/erhhung/.local/share
$ unset XDG_DATA_HOME
```
```
[76864] [INF] Starting nats-server
[76864] [INF]   Version:  2.10.0
[76864] [INF]   Git:      [not set]
[76864] [INF]   Name:     nats_development
[76864] [INF]   Node:     HhEiCaYr
[...]

PR #963, opened 2024-01-12
Fix incorrect JetStream store dir when using $XDG_DATA_HOME #962.

erhhung:
Eliminated duplicate (and incorrect) function to determine the per-user data directory based on `$XDG_DATA_HOME`.
Kept `dataParentDir()` function to maintain concept, but delegated to correct implementation `xdgShareHome()`.

### DIFF
--- a/cli/server_run_command.go
+++ b/cli/server_run_command.go
@@ -20,7 +20,6 @@ import (
 	"net/url"
 	"os"
 	"os/signal"
-	"os/user"
 	"path/filepath"
 	"runtime"
 	"strings"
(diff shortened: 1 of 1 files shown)


cli/server_run_command.go
```diff
@@ -160,21 +159,7 @@ func (c *SrvRunCmd) getRandomPort() (string, error) {
 }
 
 func (c *SrvRunCmd) dataParentDir() (string, error) {
```
ripienaar, line 161: we can probably just remove this function entirely and call the real one where needed?

erhhung: Certainly possible. I had just wanted to preserve the semantic meaning of having a "data parent dir" while delegating the implementation details to the precise meaning of `xdgShareHome()`, but it's likely never going to change since it's already a solid convention.

ripienaar:
Thanks a lot

PR #986, opened 2024-02-07
Release workflow

samuelattwood:


.github/dependabot.yml
```diff
@@ -0,0 +1,16 @@
...
+
+  # version updates: disabled
+  # security updates: enabled
+  # https://docs.github.com/en/code-security/dependabot/dependabot-security-updates/configuring-dependabot-security-updates#overriding-the-default-behavior-with-a-configuration-file
+  # ./go
+  - package-ecosystem: gomod
+    directory: /
+    schedule:
+      interval: daily
+    open-pull-requests-limit: 0
```
ripienaar, line 16: generally find dependabot just annoying and noisy, I tend to update all dependencies almost weekly on main, and for release branches we would anyway do it right before release.

ripienaar (approved):
LGTM, of course who knows with github actions.

Small comment on dependabot I just really dont want the daily annoyance

.github/dependabot.yml
samuelattwood: I can remove it if you prefer. The config I added will only open go module PRs for security updates, not for general updates.

ripienaar: Yeah if you dont mind, that'd be good

samuelattwood: I left the weekly check for github actions updates but removed the golang config. The former should be very infrequent and is easier to lose track of.

PR #1001, opened 2024-03-03
sub  command handling jsOpts

jnmoyne:
Fixes the sub command creating a JS context without taking all options (e.g. --js-domain


ripienaar:
Thank you!

PR #1025, opened 2024-03-21
password encryption UI/UX fix

zakk616:
Fixed UI/UX for password encryption

1. The word 'Reenter' is creating confusion when encrypting a password, changed it to 'Re-enter'.

2. When Re-entered password does not match entered password, typed password is being exposed.

image

changed the error line to a user-friendly sentence.


ripienaar:
Thank you!

PR #1026, opened 2024-03-22
[ADD] KV check if we need to read from stdin

ramonberrutti:
Check if we want an empty value or read from StdIn.
Added test for **kv create** and **kv update**


ripienaar:
Thanks!

PR #1031, opened 2024-04-03
adds xkey seal/open functionality to commandline

jordan-rash:
Adds the helper command to NATS cli to allow encryptions/decryption via xkeys.

```
└─❯ cat message.txt
hello

└─❯ go run . auth nkey seal message.txt ./my.key XBZ6MJBDOTEIMIEHHFWY2TBGNGOVU7QRMGHDUSOZ5OPJ2EYWCRUB32W4 --output derp.enc

└─❯ cat derp.enc
xkv1����݋i8���fܮ L^T��^K�o�^Y��$U΄�]���^^�F���r-�G

└─❯ go run . auth nkey open derp.enc ./their.key XDBAJQRNSD6OD6A6X66UWNQTQDFJWGDFJBCVKQJSZ5NA2WM7CVYA7UUP
hello
```

### DIFF
--- a/cli/auth_nkey_command.go
+++ b/cli/auth_nkey_command.go
@@ -27,13 +27,15 @@ import (
 )
 
 type authNKCommand struct {
-	keyType       string
-	pubOut        bool
-	entropySource string
-	outFile       string
-	keyFile       string
-	dataFile      string
-	signFile      string
+	keyType        string
+	pubOut         bool
+	entropySource  string
+	outFile        string
+	keyFile        string
+	dataFile       string
+	signFile       string
+	counterpartKey string
+	b64out         bool
 }
 
 func configureAuthNkeyCommand(auth commandHost) {
@@ -58,6 +60,19 @@ func configureAuthNkeyCommand(auth commandHost) {
 	nkVerify.Arg("file", "File containing the data to check").Required().ExistingFileVar(&c.dataFile)
 	nkVerify.Arg("signature", "File containing the signature").Required().ExistingFileVar(&c.signFile)
 	nkVerify.Arg("key", "The key to use for verification").Required().ExistingFileVar(&c.keyFile)
+
+	nkSeal := nk.Command("seal", "Encrypts file").Alias("encrypt").Alias("enc").Action(c.sealAction)
+	nkSeal.Arg("file", "File to sign").Required().ExistingFileVar(&c.dataFile)
+	nkSeal.Arg("key", "NKey to sign with").Required().ExistingFileVar(&c.keyFile)
+	nkSeal.Arg("receipent", "Public XKey of receipient").Required().StringVar(&c.counterpartKey)
+	nkSeal.Flag("output", "Write the encrypted data to a file").StringVar(&c.outFile)
+	nkSeal.Flag("b64", "Write base64 encoded data to stdout").Default("false").UnNegatableBoolVar(&c.b64out)
+
+	nkOpen := nk.Command("open", "Decrypts file").Alias("decrypt").Alias("dec").Action(c.openAction)
+	nkOpen.Arg("file", "File to decrypt").Required().ExistingFileVar(&c.dataFile)
+	nkOpen.Arg("key", "XKey to decrypt with").Required().ExistingFileVar(&c.keyFile)
+	nkOpen.Arg("sender", "Public XKey of sender").Required().StringVar(&c.counterpartKey)
+	nkOpen.Flag("output", "Write the decrypted data to a file").StringVar(&c.outFile)
 }
 
 func (c *authNKCommand) showAction(_ *fisk.ParseContext) error {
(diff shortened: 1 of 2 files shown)


ripienaar:
We should maybe by default nase64 or he encode the encoded file with an option to disable?

not reviewed the code yet will next week when I am back

jordan-rash:
> We should maybe by default nase64 or he encode the encoded file with an option to disable?
> 
> not reviewed the code yet will next week when I am back

Switched the sealed data output default to b64 with the ability to disable

cli/auth_nkey_command.go
```diff
@@ -58,6 +60,19 @@ func configureAuthNkeyCommand(auth commandHost) {
...
 	nkVerify.Arg("key", "The key to use for verification").Required().ExistingFileVar(&c.keyFile)
+
+	nkSeal := nk.Command("seal", "Encrypts file").Alias("encrypt").Alias("enc").Action(c.sealAction)
+	nkSeal.Arg("file", "File to sign").Required().ExistingFileVar(&c.dataFile)
+	nkSeal.Arg("key", "NKey to sign with").Required().ExistingFileVar(&c.keyFile)
+	nkSeal.Arg("receipent", "Public XKey of receipient").Required().StringVar(&c.counterpartKey)
+	nkSeal.Flag("output", "Write the encrypted data to a file").StringVar(&c.outFile)
+	nkSeal.Flag("b64", "Write base64 encoded data [Default]").Default("true").BoolVar(&c.b64out)
+
+	nkOpen := nk.Command("open", "Decrypts file").Alias("decrypt").Alias("dec").Action(c.openAction)
```
ripienaar, line 71: hmm, i wonder if `unseal` is better - making it speficially related to `seal`?

cli/auth_nkey_command.go
```diff
@@ -58,6 +60,19 @@ func configureAuthNkeyCommand(auth commandHost) {
 	nkVerify.Arg("file", "File containing the data to check").Required().ExistingFileVar(&c.dataFile)
 	nkVerify.Arg("signature", "File containing the signature").Required().ExistingFileVar(&c.signFile)
 	nkVerify.Arg("key", "The key to use for verification").Required().ExistingFileVar(&c.keyFile)
+
+	nkSeal := nk.Command("seal", "Encrypts file").Alias("encrypt").Alias("enc").Action(c.sealAction)
+	nkSeal.Arg("file", "File to sign").Required().ExistingFileVar(&c.dataFile)
+	nkSeal.Arg("key", "NKey to sign with").Required().ExistingFileVar(&c.keyFile)
+	nkSeal.Arg("receipent", "Public XKey of receipient").Required().StringVar(&c.counterpartKey)
+	nkSeal.Flag("output", "Write the encrypted data to a file").StringVar(&c.outFile)
+	nkSeal.Flag("b64", "Write base64 encoded data [Default]").Default("true").BoolVar(&c.b64out)
```
ripienaar, line 69: looks like the unseal doesnt handle base 64 encoded files so we need a similar option on the unseal side?

cli/auth_xkey_test.go
```diff
@@ -0,0 +1,125 @@
...
+	p2, err := nkeys.CreateCurveKeysWithRand(ef)
+	if err != nil {
+		t.Error("Failed to create key")
+		t.FailNow()
+	}
+	p1_seed, _ := p1.Seed()
+	p2_seed, _ := p2.Seed()
+
+	// Setup all the test files
+	p1_key, err := os.Create(filepath.Join(os.TempDir(), "p1_seed"))
```
ripienaar, line 43: better to use `t.TempDir()` rather than `os.TempDir()` which might leave stuff behind and cause future tests to maybe act weird

ripienaar (changes requested):
nice one, some small changes maybe

cli/auth_nkey_command.go
jordan-rash: I agree

jordan-rash:
was OOO last week, will get all these edits in today.  thanks!

cli/auth_xkey_test.go
jordan-rash: TIL `t.TempDir` was a thing...thanks!!

cli/auth_nkey_command.go
```diff
@@ -256,3 +272,106 @@ func (c *authNKCommand) verifyAction(_ *fisk.ParseContext) error {
 
 	return nil
 }
+func (c *authNKCommand) sealAction(_ *fisk.ParseContext) error {
+	var err error
```
ripienaar, line 276: not needed its declared 2 lines below, kp probably also not needed to be pre-declared as you can change `=` to `:=` on line 285 with no ill side effects

cli/auth_nkey_command.go
```diff
@@ -256,3 +272,106 @@ func (c *authNKCommand) verifyAction(_ *fisk.ParseContext) error {
...
+	if err != nil {
+		return err
+	}
+
+	content, err := os.ReadFile(c.dataFile)
+	if err != nil {
+		return err
+	}
+
+	if nkeys.IsValidPublicCurveKey(c.counterpartKey) {
```
ripienaar, line 298: thnk its better to change this to

```
if !nkeys.IsValidPublicCurveKey(c.counterpartKey) {
  return fmt.Errorf("some nice error")
}
```

and unindent all the rest so you handle all the validation upfront, its also just nice to not wrap the meat of a function inside a conditional - obviously a minor nit.

cli/auth_nkey_command.go
```diff
@@ -256,3 +272,106 @@ func (c *authNKCommand) verifyAction(_ *fisk.ParseContext) error {
...
+		encryptedData, err := kp.Seal(content, c.counterpartKey)
+		if err != nil {
+			return err
+		}
+		if c.useB64 {
+			encryptedData = []byte(base64.StdEncoding.EncodeToString(encryptedData))
+		}
+
+		if c.outFile != "" {
```
ripienaar, line 307: like above I'd `if c.outFile == "" { print/return }` and then have the bigger code block unindented and no need for an else then

cli/auth_nkey_command.go
```diff
@@ -256,3 +272,106 @@ func (c *authNKCommand) verifyAction(_ *fisk.ParseContext) error {
...
+			return nil
+		} else {
+			fmt.Println(string(encryptedData))
+			return nil
+		}
+	}
+
+	return errors.New("failed to seal message")
+}
+func (c *authNKCommand) openAction(_ *fisk.ParseContext) error {
```
ripienaar, line 325: rename to `unsealAction` please and then I guess the general style comments above would apply here also as they are roughly same layout.

Otherwise this is great stuff thanks for the contrib

ripienaar (changes requested):
minor nits

jordan-rash:
I think i got them all.  Thanks for the feedback!

ripienaar (approved):
LGTM! if you have rights you can merge else let me know

PR #1037, opened 2024-04-16
Support backing up operators

ripienaar:


aricart (approved):
LGTM - the only consideration is if the secret should be an actual xkey or a passphrase that generates an xkey

PR #1063, opened 2024-05-31
Move audit checks logic to internal package

mprimi:
Refactor `audit analyze` command checks into and into a separate `internal/audit` package.
Also move `archive` to `internal` and include some minor logging improvements.


ripienaar:
Does this need to be public or can it go in `internal/audit`?

mprimi:
> Does this need to be public or can it go in `internal/audit`?

I suppose that's up to you, and wether you want folks to use this stuff as library or not.
I have a weak preference for `internal`. Fewer users, fewer interrupts for me.
Same for the `archive` package

ripienaar:
Yes let’s start there and go public when we need to. We can’t go the other way once it ships :)

ripienaar (approved):
LGTM, big improvement

mprimi:
Need to set up commit signing on this new laptop... ⏳

PR #1064, opened 2024-05-31
[ADD] step-down force option

ramonberrutti:
Add the force option to ignore the current leader for:

- cluster
- stream
- consumer


ripienaar:
Thank you

PR #1090, opened 2024-07-16
New version of the bench command

jnmoyne:
Many changes, improvements and new features, including: completely different arguments and flags, using the new service framework and the new JS API.

### DIFF
--- a/cli/bench_command.go
+++ b/cli/bench_command.go
@@ -50,6 +50,7 @@ type benchCmd struct {
 	streamMaxBytesString string
 	streamMaxBytes       int64
 	ackMode              string
+	doubleAck            bool
 	batchSize            int
 	replicas             int
 	purge                bool
@@ -113,6 +114,13 @@ func configureBenchCommand(app commandHost) {
 		f.Flag("js-timeout", "Timeout for JS operations").Default("30s").DurationVar(&c.jsTimeout)
 	}
 
+	addJSConsumerFlags := func(f *fisk.CmdClause) {
+		f.Flag("consumer", "Specify the durable consumer name to use").Default(benchDefaultDurableConsumerName).StringVar(&c.consumerName)
+		f.Flag("batch", "Sets the max number of messages that can be buffered in the client").Default("500").IntVar(&c.batchSize)
+		f.Flag("acks", "Acknowledgement mode for the consumer").Default(benchAckModeExplicit).EnumVar(&c.ackMode, benchAckModeExplicit, benchAckModeNone, benchAckModeAll)
+		f.Flag("doubleack", "Synchronously acknowledge messages, waiting for a reply from the server").Default("false").BoolVar(&c.doubleAck)
+	}
+
 	addJSPubFlags := func(f *fisk.CmdClause) {
 		f.Flag("create", "Create or update the stream first").UnNegatableBoolVar(&c.createStream)
 		f.Flag("storage", "JetStream storage (memory/file) for the \"benchstream\" stream").Default("file").EnumVar(&c.storage, "memory", "file")
@@ -120,6 +128,7 @@ func configureBenchCommand(app commandHost) {
 		f.Flag("maxbytes", "The maximum size of the stream or KV bucket in bytes").Default("1GB").StringVar(&c.streamMaxBytesString)
 		f.Flag("dedup", "Sets a message id in the header to use JS Publish de-duplication").Default("false").UnNegatableBoolVar(&c.deDuplication)
 		f.Flag("dedupwindow", "Sets the duration of the stream's deduplication functionality").Default("2m").DurationVar(&c.deDuplicationWindow)
+		f.Flag("batch", "The number of asynchronous JS publish calls before waiting for all the publish acknowledgements (set to 1 for synchronous)").Default("500").IntVar(&c.batchSize)
 	}
 
 	addKVPutFlags := func(f *fisk.CmdClause) {
@@ -164,7 +173,6 @@ func configureBenchCommand(app commandHost) {
 
 	jspub := jsCommand.Command("pub", "Publish JetStream messages").Action(c.jspubAction)
 	jspub.Arg("subject", "Subject to use for the benchmark").Required().StringVar(&c.subject)
-	jspub.Flag("batch", "The number of asynchronous JS publish calls before waiting for all the publish acknowledgements (set to 1 for synchronous)").Default("500").IntVar(&c.batchSize)
 	addPubFlags(jspub)
 	addJSPubFlags(jspub)
 
@@ -172,14 +180,10 @@ func configureBenchCommand(app commandHost) {
 	jsOrdered.Flag("batch", "Sets the max number of messages that can be buffered in the client").Default("500").IntVar(&c.batchSize)
 
 	jsConsume := jsCommand.Command("consume", "Consume JetStream messages from a durable consumer using a callback").Action(c.jsConsumeAction)
-	jsConsume.Flag("consumer", "Specify the durable consumer name to use").Default(benchDefaultDurableConsumerName).StringVar(&c.consumerName)
-	jsConsume.Flag("batch", "Sets the max number of messages that can be buffered in the client").Default("500").IntVar(&c.batchSize)
-	jsConsume.Flag("acks", "Acknowledgement mode for the consumer").Default(benchAckModeExplicit).EnumVar(&c.ackMode, benchAckModeExplicit, benchAckModeNone, benchAckModeAll)
+	addJSConsumerFlags(jsConsume)
 
 	jsFetch := jsCommand.Command("fetch", "Consume JetStream messages from a durable consumer using fetch").Action(c.jsFetchAction)
-	jsFetch.Flag("consumer", "Specify the durable consumer name to use").Default(benchDefaultDurableConsumerName).StringVar(&c.consumerName)
-	jsFetch.Flag("batch", "Sets the fetch batch size").Default("500").IntVar(&c.batchSize)
-	jsFetch.Flag("acks", "Acknowledgement mode for the consumer").Default(benchAckModeExplicit).EnumVar(&c.ackMode, benchAckModeExplicit, benchAckModeNone, benchAckModeAll)
+	addJSConsumerFlags(jsFetch)
 
 	kvCommand := benchCommand.Command("kv", "KV benchmark operations")
 	addCommonFlags(kvCommand)
@@ -207,12 +211,15 @@ func configureBenchCommand(app commandHost) {
 	oldJSPush.Flag("consumer", "Specify the durable consumer name to use").Default(benchDefaultDurableConsumerName).StringVar(&c.consumerName)
 	oldJSPush.Flag("maxacks", "Sets the max ack pending value, adjusts for the number of clients").Default("500").IntVar(&c.batchSize)
 	oldJSPush.Flag("ack", "Uses explicit message acknowledgement or not for the consumer").Default("true").BoolVar(&c.ack)
+	oldJSPush.Flag("doubleack", "Synchronously acknowledge messages, waiting for a reply from the server").Default("false").BoolVar(&c.doubleAck)
 
 	oldJSPull := oldJSCommand.Command("pull", "Consume JetStream messages from a consumer using an old JS API's durable pull consumer").Action(c.oldjsPullAction)
 	oldJSPull.Arg("subject", "Subject to use for the benchmark").Required().StringVar(&c.subject)
 	oldJSPull.Flag("consumer", "Specify the durable consumer name to use").Default(benchDefaultDurableConsumerName).StringVar(&c.consumerName)
 	oldJSPull.Flag("batch", "Sets the fetch size for the consumer").Default("500").IntVar(&c.batchSize)
 	oldJSPull.Flag("ack", "Uses explicit message acknowledgement or not for the consumer").Default("true").BoolVar(&c.ack)
+	oldJSPull.Flag("doubleack", "Synchronously acknowledge messages, waiting for a reply from the server").Default("false").BoolVar(&c.doubleAck)
+
 }
 
 func init() {
@@ -340,6 +347,7 @@ func (c *benchCmd) generateBanner(benchType string) string {
 		benchTypeLabel = "JetStream durable consumer (callback)"
 		argnvps = append(argnvps, nvp{"consumer", c.consumerName})
 		argnvps = append(argnvps, nvp{"acks", c.ackMode})
+		argnvps = append(argnvps, nvp{"double-acked", f(c.doubleAck)})
 		argnvps = append(argnvps, nvp{"batch", f(c.batchSize)})
 		jsAttributes()
 		streamOrBucketAttribues()
@@ -347,6 +355,7 @@ func (c *benchCmd) generateBanner(benchType string) string {
 		benchTypeLabel = "JetStream durable consumer (fetch)"
 		argnvps = append(argnvps, nvp{"consumer", c.consumerName})
 		argnvps = append(argnvps, nvp{"acks", c.ackMode})
+		argnvps = append(argnvps, nvp{"double-acked", f(c.doubleAck)})
 		argnvps = append(argnvps, nvp{"batch", f(c.batchSize)})
 		jsAttributes()
 		streamOrBucketAttribues()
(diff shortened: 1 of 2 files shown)


cli/bench_command.go
```diff
@@ -66,305 +62,540 @@ type benchCmd struct {
 	deDuplicationWindow  time.Duration
 	retries              int
 	retriesUsed          bool
+	ack                  bool
 }
 
 const (
-	DefaultDurableConsumerName string = "natscli-bench"
+	DefaultDurableConsumerName string = "nats-bench"
```
ripienaar, line 69: be good to prefix all of these with Bench and then not export them

cli/bench_command.go
```diff
@@ -66,305 +62,540 @@ type benchCmd struct {
...
 
 	benchHelp := `
-Core NATS publish and subscribe:
+Core NATS publish:
 
-  nats bench benchsubject --pub 1 --sub 10
+  nats bench pub benchsubject
 
-Request reply with queue group:
+Core NATS subscribe:
```
ripienaar, line 100: nice its split into commands but I think we should focus on the help on each sub command keeping with the overall style of the CLI being more list with short description above.  If the individual commands are well documented and have good descriptions it largely remove the need for this

```
Subcommands:
  stream add       Create a new Stream
  stream ls        List all known Streams
  stream report    Reports on Stream statistics
  stream find      Finds streams matching certain criteria
  stream info      Stream information
```

Then expand each commands help individually with more details and perhaps even add some examples into `nats cheat`

cli/bench_command.go
```diff
@@ -66,305 +62,540 @@ type benchCmd struct {
...
-	bench.Flag("multisubject", "Multi-subject mode, each message is published on a subject that includes the publisher's message sequence number as a token").UnNegatableBoolVar(&c.multiSubject)
-	bench.Flag("multisubjectmax", "The maximum number of subjects to use in multi-subject mode (0 means no max)").Default("100000").IntVar(&c.multiSubjectMax)
-	bench.Flag("retries", "The maximum number of retries in JS operations").Default("3").IntVar(&c.retries)
-	bench.Flag("dedup", "Sets a message id in the header to use JS Publish de-duplication").Default("false").UnNegatableBoolVar(&c.deDuplication)
-	bench.Flag("dedupwindow", "Sets the duration of the stream's deduplication functionality").Default("2m").DurationVar(&c.deDuplicationWindow)
+
+	addCommonFlags := func(f *fisk.CmdClause) {
+		f.Flag("clients", "Number of concurrent clients").Default("1").IntVar(&c.numClients)
+		f.Flag("msgs", "Number of messages to publish or subscribe to").Default("100000").IntVar(&c.numMsg)
+		f.Flag("no-progress", "Disable progress bar").UnNegatableBoolVar(&c.noProgress)
```
ripienaar, line 142: elsewhere we add the `progress` flag as a bool and make it default to `false` which achieves this and ensure consistently rendering the inverse options

cli/bench_command.go
```diff
@@ -66,305 +62,540 @@ type benchCmd struct {
...
+		f.Flag("csv", "Save benchmark data to CSV file").StringVar(&c.csvFile)
+		f.Flag("size", "Size of the test messages").Default("128").StringVar(&c.msgSizeString)
+		// TODO: support randomized payload data
+	}
+
+	addPubFlags := func(f *fisk.CmdClause) {
+		f.Flag("multisubject", "Multi-subject mode, each message is published on a subject that includes the publisher's message sequence number as a token").UnNegatableBoolVar(&c.multiSubject)
+		f.Flag("multisubjectmax", "The maximum number of subjects to use in multi-subject mode (0 means no max)").Default("100000").IntVar(&c.multiSubjectMax)
+		f.Flag("sleep", "Sleep for the specified interval after publishing each message").Default("0s").DurationVar(&c.sleep)
```
ripienaar, line 151: helps for these to do something like `PlaceHolder("DURATION")`, in fact using place holder can really improve the overall output of help - I dont use it nearly enough

cli/bench_command.go
```diff
@@ -66,305 +62,540 @@ type benchCmd struct {
...
+		f.Flag("storage", "JetStream storage (memory/file) for the \"benchstream\" stream").Default("file").EnumVar(&c.storage, "memory", "file")
+		f.Flag("replicas", "Number of stream replicas for the \"benchstream\" stream").Default("1").IntVar(&c.replicas)
+		f.Flag("maxbytes", "The maximum size of the stream or KV bucket in bytes").Default("1GB").StringVar(&c.streamMaxBytesString)
+		f.Flag("retries", "The maximum number of retries in JS operations").Default("3").IntVar(&c.retries)
+		f.Flag("dedup", "Sets a message id in the header to use JS Publish de-duplication").Default("false").UnNegatableBoolVar(&c.deDuplication)
+		f.Flag("dedupwindow", "Sets the duration of the stream's deduplication functionality").Default("2m").DurationVar(&c.deDuplicationWindow)
+	}
+
+	addKVFlags := func(f *fisk.CmdClause) {
+		f.Flag("purge", "Purge the stream before running").UnNegatableBoolVar(&c.purge)
```
ripienaar, line 171: be good to use kv terminology here so bucket not stream etc

cli/bench_command.go
```diff
@@ -66,305 +62,540 @@ type benchCmd struct {
...
+
+	jsCommand := benchCommand.Command("js", "JetStream benchmark commands")
+
+	jspub := jsCommand.Command("pub", "Publish JetStream messages").Action(c.jspubAction)
+	jspub.Arg("subject", "Subject to use for the benchmark").Required().StringVar(&c.subject)
+	jspub.Flag("batch", "The number of asynchronous JS publish calls before waiting for all the publish acknowledgements (set to 1 for synchronous)").Default("100").IntVar(&c.batchSize)
+	addCommonFlags(jspub)
+	addPubFlags(jspub)
+	addJSCommonFlags(jspub)
+	addJSPubFlags(jspub)
```
ripienaar, line 215: adding js flags - and all the many other flags - seems a bit regressive and ends up with another hugely complex command not much different from what we had. Is it not possible to do the js stuff under the js sub command?

cli/bench_command.go
```diff
@@ -66,305 +62,540 @@ type benchCmd struct {
...
+
+	jsOrdered := jsCommand.Command("ordered", "Consume JetStream messages from a consumer using an ephemeral ordered consumer").Action(c.jsOrderedAction)
+	jsOrdered.Flag("consumer", "Specify the durable consumer name to use").Default(DefaultDurableConsumerName).StringVar(&c.consumerName)
```
ripienaar, line 218: above you say '...using an ephemeral' here you say durable consumer name - i dont think ordered supports durable?

cli/bench_command.go
```diff
@@ -66,305 +62,540 @@ type benchCmd struct {
...
+	addJSCommonFlags(kvput)
+	addKVFlags(kvput)
+
+	kvget := kvCommand.Command("get", "Get messages from a KV bucket").Action(c.kvGetAction)
+	kvget.Arg("bucket", "The bucket to use for the benchmark").Default(DefaultBucketName).StringVar(&c.bucketName)
+	kvget.Flag("sleep", "Sleep for the specified interval after getting each message").Default("0s").DurationVar(&c.sleep)
+	addCommonFlags(kvget)
+	addJSCommonFlags(kvget)
+
+	oldJSCommand := benchCommand.Command("oldjs", "JetStream benchmark commands using the old JS API")
```
ripienaar, line 255: I think we can just remove this now

cli/bench_command.go
```diff
@@ -411,232 +665,1267 @@ func (c *benchCmd) bench(_ *fisk.ParseContext) error {
...
-	var err error
+	// create the stream for the benchmark (and purge it)
+	nc, err := nats.Connect(opts().Config.ServerURL(), natsOpts()...)
+	if err != nil {
+		log.Fatalf("NATS connection failed: %v", err)
+	}
 
-	errBytes := []byte("error")
-	minusByte := byte('-')
+	ctx, cancel := context.WithTimeout(context.Background(), c.jsTimeout)
```
ripienaar, line 737: there is a top level ctx defined please use that instead of context.Background(), today that is also context.Background but later we might want to install a overall signal handler or something.

cli/bench_command.go
```diff
@@ -411,232 +665,1267 @@ func (c *benchCmd) bench(_ *fisk.ParseContext) error {
...
+		ack = nats.AckExplicitPolicy
+	}
+
+	if c.consumerName == DefaultDurableConsumerName {
+		_, err = js.AddConsumer(c.streamName, &nats.ConsumerConfig{
+			Durable:       c.consumerName,
+			DeliverPolicy: nats.DeliverAllPolicy,
+			AckPolicy:     ack,
+			ReplayPolicy:  nats.ReplayInstantPolicy,
+			MaxAckPending: func(a int) int {
```
ripienaar, line 1291: better pull this into a helper like the getSubscribeSubject()

cli/bench_command.go
```diff
@@ -411,232 +665,1267 @@ func (c *benchCmd) bench(_ *fisk.ParseContext) error {
...
+
+	err = c.printResults(bm)
+	if err != nil {
+		return err
+	}
+
+	return nil
+}
+
+func getSubscribeSubject(c *benchCmd) string {
```
ripienaar, line 1344: please attach all these to the `banchCmd` so we dont polute the global function space so much with undocumented and unclearly named functions

ripienaar (changes requested):
this is a huge improvement thanks for doing this.

Left some minor UX comments but didnt go to deep into the bench code you know that much better than me

cli/bench_command.go
jnmoyne: It's a durable consumer (except for ordered of course) but it's deleted at the end of a successful bench run, and also has a pretty short idle timeout, so it only really exists while the bench is running.
So you would rather the name was 'bench-natscli' ?

ripienaar: I mean the variable names :) since these clutter up the the entire cli package it would be good to have them be like `benchDefaultDurableConsumerName`

cli/bench_command.go
jnmoyne: Reworked the help and updated the cheat sheet file as well. Indeed it is now much better with the sub commands now there's no need to provide examples in the in-line help (but they are there in the cheats sheet file).

cli/bench_command.go
jnmoyne: ok changed to progressBar

cli/bench_command.go
jnmoyne: Done

cli/bench_command.go
jnmoyne: Good catch, done.

cli/bench_command.go
jnmoyne: Good idea, moved common flags under the command for the js, service and kv (and oldjs) commands

cli/bench_command.go
jnmoyne: Good catch, flag removed.

cli/bench_command.go
jnmoyne: I thought about it and I did this oldjs because I thought it would be useful to compare quickly the difference in speed (if any) between the old and the new JS API. So my vote is to keep it there for the moment and could always remove it later.

As a half-way compromise what I just did is mark the 'oldjs' command as hidden, so it's still there but you won't see it :).

cli/bench_command.go
jnmoyne: Done

cli/bench_command.go
jnmoyne: Actually replaced it with min since we (finally!:)) have that in Go now.

cli/bench_command.go
jnmoyne: Sure, done.

cli/bench_command.go
jnmoyne: Ah ok, got ya, done.

cli/bench_command.go
```diff
@@ -66,305 +62,497 @@ type benchCmd struct {
...
 	retries              int
 	retriesUsed          bool
+	ack                  bool
 }
 
 const (
-	DefaultDurableConsumerName string = "natscli-bench"
-	DefaultStreamName          string = "benchstream"
-	DefaultBucketName          string = "benchbucket"
+	BenchDefaultDurableConsumerName string = "nats-bench"
```
ripienaar, line 69: please unexport all these, just `benchDefaultDurableConsumerName` not `BenchDefaultDurableConsumerName`

cli/bench_command.go
```diff
@@ -66,305 +62,497 @@ type benchCmd struct {
...
+	BenchTypeJSConsume              string = "jsconsume"
+	BenchTypeJSFetch                string = "jsfetch"
+	BenchTypeOldJSOrdered           string = "oldjsordered"
+	BenchTypeOldJSPush              string = "oldjspush"
+	BenchTypeOldJSPull              string = "oldjspull"
+	BenchTypeKVPut                  string = "kvput"
+	BenchTypeKVGet                  string = "kvget"
+	AckModeNone                     string = "none"
+	AckModeAll                      string = "all"
+	AckModeExplicit                 string = "explicit"
```
ripienaar, line 89: no biggie but `string` in all of these are not needed.

cli/bench_command.go
```diff
@@ -66,305 +62,497 @@ type benchCmd struct {
...
+}
+
+func (c *benchCmd) storageType() jetstream.StorageType {
+	switch c.storage {
+	case "file":
+		return jetstream.FileStorage
+	case "memory":
+		return jetstream.MemoryStorage
+	default:
+		{
```
ripienaar, line 457: dont need these `{}` around - but also I dont think this default case it needed here since the variable is set by Fisk using a enumvar it can only ever be on or the other.

But even if we keep this we should default to file not memory - since thats what the CLI flags are set to do

cli/bench_command.go
```diff
@@ -411,72 +622,841 @@ func (c *benchCmd) bench(_ *fisk.ParseContext) error {
...
+	log.Println(banner)
+	bm := bench.NewBenchmark("NATS", 0, c.numClients)
+	benchId := strconv.FormatInt(time.Now().UnixMilli(), 16)
+	startwg := &sync.WaitGroup{}
+	donewg := &sync.WaitGroup{}
+
+	// create the stream for the benchmark (and purge it)
+	nc, err := nats.Connect(opts().Config.ServerURL(), natsOpts()...)
+	if err != nil {
+		log.Fatalf("NATS connection failed: %v", err)
```
ripienaar, line 691: I'd prefer we dont use `log.Fatalf` but instead return errors so that the cli processor can handle the exit and printing of errors etc

cli can be used as a package so a package dying without giving callers a way to deal with that isnt super nice

with these being many go routines etc I can imagine getting this working is quite annoying but you could make a errors channel on `c` and check that at the call spots

cli/bench_command.go
```diff
@@ -411,72 +622,841 @@ func (c *benchCmd) bench(_ *fisk.ParseContext) error {
...
+
+	err = c.printResults(bm)
+	if err != nil {
+		return err
+	}
+
+	return nil
+}
+
+func coreNATSPublisher(c benchCmd, nc *nats.Conn, progress *uiprogress.Bar, msg []byte, numMsg int, offset int) {
```
ripienaar, line 1295: some more funds that should move to `*c` here

ripienaar (changes requested):
looking great

cli/bench_command.go
```diff
@@ -66,305 +62,497 @@ type benchCmd struct {
...
+	oldJSPull.Flag("ack", "Uses explicit message acknowledgement or not for the consumer").Default("true").BoolVar(&c.ack)
+}
 
-  nats bench benchsubject --js --purge --pub 1
+func init() {
+	registerCommand("bench", 2, configureBenchCommand)
+}
 
-JetStream ordered ephemeral consumers:
+func getJS(nc *nats.Conn) (jetstream.JetStream, error) {
```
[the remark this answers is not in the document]
jnmoyne: Moved all functions to benchCmd pointer receivers

cli/bench_command.go
jnmoyne: changed all my consts to not be exported

cli/bench_command.go
```diff
@@ -66,305 +62,497 @@ type benchCmd struct {
...
+	banner := fmt.Sprintf("Starting %s benchmark [", benchTypeLabel)
+
+	var joinBuffer []string
+
+	for _, v := range argnvps {
+		joinBuffer = append(joinBuffer, v.name+"="+v.value)
 	}
-	if c.js && c.numSubs > 0 && !c.pull {
-		log.Print("JetStream ephemeral ordered push consumer mode, subscribers will not acknowledge the consumption of messages")
+	banner += strings.Join(joinBuffer, ", ") + "]"
```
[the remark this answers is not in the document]
jnmoyne: We could think of some kind of better way down the road, in the mean time I did sort them (which as it happens gives a pretty good order :))

cli/bench_command.go
jnmoyne: Cleaned up

cli/bench_command.go
jnmoyne: Yes, overkill for the need, simplified to an if statement

cli/bench_command.go
jnmoyne: So basically re-did error handling in general everywhere for the bench command, which was indeed quite annoying and a fair bit of work, especially dealing with the go routines but now the code is completely free of log.Fatalf and returns the error all the way back up but the errors are better now (homogenized, wrapped, etc...) for it. Note that because each of the go routines could experience an error and only one can be returned it will print all of the errors for all the go routines but only return the first one back up.

cli/bench_command.go
jnmoyne: Moved all functions to benchCmd pointer receivers

ripienaar (approved):
This is looking great, I am happy to merge it and handle any issues in followups so merge whenever you are ready @jnmoyne

PR #1096, opened 2024-07-23
Add whitespace checks to audit analyze

ReubenMathew:
Signed-off-by: reubenninan <reuben@nats.io>

### DIFF
--- a/internal/audit/cluster_checks.go
+++ b/internal/audit/cluster_checks.go
@@ -16,11 +16,14 @@ package audit
 import (
 	"errors"
 	"fmt"
+	"reflect"
+	"sort"
+	"strings"
+
 	"github.com/dustin/go-humanize"
 	"github.com/nats-io/nats-server/v2/server"
 	"github.com/nats-io/natscli/internal/archive"
-	"reflect"
-	"sort"
+	"golang.org/x/exp/maps"
 )
 
 // checkClusterMemoryUsageOutliers creates a parametrized check to verify the memory usage of any given node in a
@@ -213,4 +216,58 @@ func makeCheckClusterHighHAAssets(haAssetsThreshold int) checkFunc {
 
 		return Pass, nil
 	}
+}
+
+func checkClusterNamesForWhitespace(reader *archive.Reader, examples *ExamplesCollection) (Outcome, error) {
+
+	for _, clusterName := range reader.GetClusterNames() {
+		if strings.Contains(clusterName, " ") {
+			examples.add("Cluster: %s", clusterName)
+		}
+	}
+
+	if examples.Count() > 0 {
+		logCritical("Found %d clusters with names containing whitespace", examples.Count())
+		return Fail, nil
+	}
+
+	return Pass, nil
+}
+
+func checkLeafnodeServerNamesForWhitespace(r *archive.Reader, examples *ExamplesCollection) (Outcome, error) {
+
+	for _, clusterName := range r.GetClusterNames() {
+		clusterTag := archive.TagCluster(clusterName)
+
+		leafnodesWithWhitespace := map[string]struct{}{}
+
+		for _, serverName := range r.GetClusterServerNames(clusterName) {
+			serverTag := archive.TagServer(serverName)
+
+			var serverLeafz server.Leafz
+			err := r.Load(&serverLeafz, clusterTag, serverTag, archive.TagServerLeafs())
+			if err != nil {
+				logWarning("Artifact 'LEAFZ' is missing for server %s", serverName)
+				continue
+			}
+
+			for _, leaf := range serverLeafz.Leafs {
+				// check if leafnode name contains whitespace
+				if strings.Contains(leaf.Name, " ") {
+					leafnodesWithWhitespace[leaf.Name] = struct{}{}
+				}
+			}
+		}
+
+		if len(leafnodesWithWhitespace) > 0 {
+			examples.add("Cluster %s: %v", clusterName, maps.Keys(leafnodesWithWhitespace))
+		}
+	}
+
+	if examples.Count() > 0 {
+		logCritical("Found %d clusters with leafnode names containing whitespace", examples.Count())
+		return Fail, nil
+	}
+
+	return Pass, nil
 }
--- a/internal/audit/server_checks.go
+++ b/internal/audit/server_checks.go
@@ -16,6 +16,8 @@ package audit
 import (
 	"errors"
 	"fmt"
+	"strings"
+
 	"github.com/dustin/go-humanize"
 	"github.com/nats-io/nats-server/v2/server"
 	"github.com/nats-io/natscli/internal/archive"
@@ -229,4 +231,34 @@ func makeCheckServerResourceLimits(memoryUsageThreshold, storeUsageThreshold flo
 
 		return Pass, nil
 	}
+}
+
+func checkJetStreamDomainsForWhitespace(r *archive.Reader, examples *ExamplesCollection) (Outcome, error) {
+
+	for _, clusterName := range r.GetClusterNames() {
+		clusterTag := archive.TagCluster(clusterName)
+
+		for _, serverName := range r.GetClusterServerNames(clusterName) {
+			serverTag := archive.TagServer(serverName)
+
+			var serverJsz server.JSInfo
+			err := r.Load(&serverJsz, clusterTag, serverTag, archive.TagServerJetStream())
+			if err != nil {
+				logWarning("Artifact 'JSZ' is missing for server %s", serverName)
+				continue
+			}
+
+			// check if jetstream domain contains whitespace
+			if strings.Contains(serverJsz.Config.Domain, " ") {
+				examples.add("Cluster %s Server %s Domain %s", clusterName, serverName, serverJsz.Config.Domain)
+			}
+		}
+	}
+
+	if examples.Count() > 0 {
+		logCritical("Found %d servers with JetStream domains containing whitespace", examples.Count())
+		return Fail, nil
+	}
+
+	return Pass, nil
 }
--- a/internal/audit/checks.go
+++ b/internal/audit/checks.go
@@ -156,6 +156,21 @@ func GetDefaultChecks() []Check {
 			Description: "Verify that all nodes part of the meta group agree on the meta cluster leader",
 			fun:         checkMetaClusterLeader,
 		},
+		{
+			Name:        "Whitespace in leafnode server names",
+			Description: "Verify that no leafnode contains whitespace in its name",
+			fun:         checkLeafnodeServerNamesForWhitespace,
+		},
+		{
+			Name:        "Whitespace in JetStream domains",
+			Description: "Verify that no JetStream server is configured with whitespace in its domain",
+			fun:         checkJetStreamDomainsForWhitespace,
+		},
+		{
+			Name:        "Whitespace in cluster name",
+			Description: "Verify that no cluster name contains whitespace",
+			fun:         checkClusterNamesForWhitespace,
+		},
 	}
 }
 


internal/audit/server_checks.go
```diff
@@ -230,3 +230,48 @@ func makeCheckServerResourceLimits(memoryUsageThreshold, storeUsageThreshold flo
 		return Pass, nil
 	}
 }
+
+func checkServerLeafnodeNames() checkFunc {
+	return func(r *archive.Reader, examples *ExamplesCollection) (Outcome, error) {
+
+		checkIfNameContainsSpaces := func(name string) bool {
```
mprimi, line 237: this should probably check for the entire class of unicode 'whitespace', not just a space.

internal/audit/server_checks.go
```diff
@@ -230,3 +230,48 @@ func makeCheckServerResourceLimits(memoryUsageThreshold, storeUsageThreshold flo
...
+		for _, clusterName := range r.GetClusterNames() {
+			clusterTag := archive.TagCluster(clusterName)
+
+			for _, serverName := range r.GetClusterServerNames(clusterName) {
+				serverTag := archive.TagServer(serverName)
+
+				var serverLeafz server.Leafz
+				err := r.Load(&serverLeafz, clusterTag, serverTag, archive.TagServerLeafs())
+				if err != nil {
+					return Skipped, fmt.Errorf("failed to load Leafz for server %s: %w", serverName, err)
```
mprimi, line 255: Returning here skips the entire check if one of the servers is missing `leafz`, should probably just print a warning and continue

internal/audit/server_checks.go
```diff
@@ -230,3 +230,48 @@ func makeCheckServerResourceLimits(memoryUsageThreshold, storeUsageThreshold flo
...
+
+				for _, leaf := range serverLeafz.Leafs {
+					// check if leaf name contains any spaces
+					if checkIfNameContainsSpaces(leaf.Name) {
+						examples.add("%s/%s: %s", clusterName, serverName, leaf.Name)
+					}
+				}
+
+			}
+			if examples.Count() > 0 {
```
mprimi, line 270: I think you should move this down a few lines otherwise it stops at the first cluster

internal/audit/server_checks.go
```diff
@@ -230,3 +230,48 @@ func makeCheckServerResourceLimits(memoryUsageThreshold, storeUsageThreshold flo
...
+				logCritical("Found %d instances of servers with leafnode names containing spaces", examples.Count())
+				return Fail, nil
+			}
+		}
+		return Pass, nil
```
mprimi, line 275: To make this better I'd suggest creating a set of invalid leafnode names per cluster, adding that as a single example.

i.e. above:
```
if contains whitespace
  invalidLeafnodes[leafnodeName] = struct{}
```

at the end of the cluster loop:
```
if len(invalidLeafnodes) > 0 {
  examples.add("%s: %s", clusterName, invalidLeafnodes.keys())
}
```

mprimi (changes requested):

mprimi:
@ripienaar other than checking for invalid **connected** leaf nodes, should we be checking anything else?
I'm thinking:
 - Server name
 - Cluster name
 - Consumer name
 - ...

ripienaar:
cluster and domain names would be great, consumers/stream names would fail to create already

internal/audit/server_checks.go
ripienaar: while I agree, for the purpose of the case at hand only actual space would be a problem - other unicode space-like characters are fine

internal/audit/server_checks.go
```diff
@@ -230,3 +230,48 @@ func makeCheckServerResourceLimits(memoryUsageThreshold, storeUsageThreshold flo
...
+func checkServerLeafnodeNames() checkFunc {
+	return func(r *archive.Reader, examples *ExamplesCollection) (Outcome, error) {
+
+		checkIfNameContainsSpaces := func(name string) bool {
+			for _, c := range name {
+				if c == ' ' {
+					return true
+				}
+			}
+			return false
```
ripienaar, line 243: `strings. Contains()`

ripienaar:
No pressure but if yous can get this one in today I can pull it into the maintenance release also

internal/audit/cluster_checks.go
```diff
@@ -214,3 +218,58 @@ func makeCheckClusterHighHAAssets(haAssetsThreshold int) checkFunc {
 		return Pass, nil
 	}
 }
+
+func checkClusterNamesForWhitespace(reader *archive.Reader, examples *ExamplesCollection) (Outcome, error) {
+
+	for _, clusterName := range reader.GetClusterNames() {
+		if strings.ContainsFunc(clusterName, unicode.IsSpace) {
+			examples.add("Cluster name '%s' contains whitespace", clusterName)
```
mprimi, line 226: This is an example printed after the check name, so 'contains whitespace is redundant' it should just be:

`examples.add("Cluster: '%s', clusterName)`

internal/audit/cluster_checks.go
```diff
@@ -214,3 +218,58 @@ func makeCheckClusterHighHAAssets(haAssetsThreshold int) checkFunc {
...
+		}
+	}
+
+	if examples.Count() > 0 {
+		logCritical("Found %d clusters with names containing spaces", examples.Count())
```
mprimi, line 231: Should be "containing whitespace" (slightly different than "containing spaces")

internal/audit/cluster_checks.go
```diff
@@ -214,3 +218,58 @@ func makeCheckClusterHighHAAssets(haAssetsThreshold int) checkFunc {
...
+
+	return Pass, nil
+}
+
+func checkLeafnodeServerNamesForWhitespace(r *archive.Reader, examples *ExamplesCollection) (Outcome, error) {
+
+	for _, clusterName := range r.GetClusterNames() {
+		clusterTag := archive.TagCluster(clusterName)
+
+		invalidLeafnodes := map[string]struct{}{}
```
mprimi, line 243: Rename to `leafNodesWithWhitespace` or maybe `leafNodeNames`?
"Invalid" is a bit misleading.

internal/audit/cluster_checks.go
```diff
@@ -214,3 +218,58 @@ func makeCheckClusterHighHAAssets(haAssetsThreshold int) checkFunc {
...
+			serverTag := archive.TagServer(serverName)
+
+			var serverLeafz server.Leafz
+			err := r.Load(&serverLeafz, clusterTag, serverTag, archive.TagServerLeafs())
+			if err != nil {
+				logWarning("Artifact 'LEAFZ' is missing for server %s", serverName)
+				continue
+			}
+
+			// check if leafnode name contains any spaces
```
mprimi, line 255: Duplicate comment

internal/audit/cluster_checks.go
```diff
@@ -214,3 +218,58 @@ func makeCheckClusterHighHAAssets(haAssetsThreshold int) checkFunc {
...
+			for _, leaf := range serverLeafz.Leafs {
+				// check if leaf name contains any spaces
+				if strings.ContainsFunc(leaf.Name, unicode.IsSpace) {
+					invalidLeafnodes[leaf.Name] = struct{}{}
+				}
+			}
+		}
+
+		if len(invalidLeafnodes) > 0 {
+			examples.add("Detected leafnodes with whitespaces in cluster %s: %v", clusterName, maps.Keys(invalidLeafnodes))
```
mprimi, line 265: Same here, this is an example, not a printf. So maybe:
  `"Cluster %s: %v", clusterName, map.Keys`

internal/audit/cluster_checks.go
mprimi: Also should say "whitespace"

internal/audit/server_checks.go
```diff
@@ -230,3 +233,33 @@ func makeCheckServerResourceLimits(memoryUsageThreshold, storeUsageThreshold flo
...
+			serverTag := archive.TagServer(serverName)
+
+			var serverJsz server.JSInfo
+			err := r.Load(&serverJsz, clusterTag, serverTag, archive.TagServerJetStream())
+			if err != nil {
+				logWarning("Artifact 'JSZ' is missing for server %s", serverName)
+				continue
+			}
+
+			// check if jetstream domain name contains any spaces
```
mprimi, line 252: "...contains whitespace"

internal/audit/server_checks.go
```diff
@@ -230,3 +233,33 @@ func makeCheckServerResourceLimits(memoryUsageThreshold, storeUsageThreshold flo
...
+			if strings.ContainsFunc(serverJsz.Config.Domain, unicode.IsSpace) {
+				examples.add("Cluster %s Server %s has a JetStream domain containing whitespaces: %s", clusterName, serverName, serverJsz.Config.Domain)
```
mprimi, line 254: `"Cluster %s server %s domain %s"  clusterName, serverName, serverJsz.Config.Domain`

mprimi (changes requested):
Very minor comments on comments names and messages. Otherwise looks good.

PR description could also use a refresh

internal/audit/server_checks.go
mprimi: I was thinking of newlines, tabs, etc. `unicode.IsSpace` does this.

mprimi (approved):
LGTM. Did you give it a whirl against NGS by any chance?

ReubenMathew:
> LGTM. Did you give it a whirl against NGS by any chance?

Yes, against `ngstest`. It passes.

ripienaar:
Can yous please setup commit signing?

internal/audit/cluster_checks.go
```diff
@@ -214,3 +218,57 @@ func makeCheckClusterHighHAAssets(haAssetsThreshold int) checkFunc {
 		return Pass, nil
 	}
 }
+
+func checkClusterNamesForWhitespace(reader *archive.Reader, examples *ExamplesCollection) (Outcome, error) {
+
+	for _, clusterName := range reader.GetClusterNames() {
+		if strings.ContainsFunc(clusterName, unicode.IsSpace) {
```
ripienaar, line 225: It's worth stating again the problem we're facing is just actual space, not things that look like space but actual space.  Space has significance in protocol parsing and so the message is parsed incorrectly.  Things that look like space is not significant.

In 2.11 we will probably add constraints to make these names be `restricted-term` as defined by ADR-6

ReubenMathew: Ah, I see. Just spaces? What about newlines, for example?

ReubenMathew:
@ripienaar commit signed, checking for spaces only.

internal/audit/cluster_checks.go
ripienaar: I doubt it would be possible to enter a newline config parser would probably fail or chop it

ripienaar: so bad, ok we should probably check for that ugh - maybe we should just check for anything thats not `restricted-term`?

ripienaar: cc @wallyqs just in case there's some more restrictions we should put in as newline for sure would cause protocol issues

ReubenMathew: @ripienaar, are there any "whitespace" characters filtered out by `unicode.IsSpace` that could/should be used in a server name?

```
// IsSpace reports whether the rune is a space character as defined
// by Unicode's White Space property; in the Latin-1 space
// this is
//
//	'\t', '\n', '\v', '\f', '\r', ' ', U+0085 (NEL), U+00A0 (NBSP).
//
// Other definitions of spacing characters are set by category
// Z and property Pattern_White_Space.
func IsSpace(r rune) bool {
```

ripienaar: Probsbly not but the reason I am trying to be minimalist here is that our initial aim is to not break LTS users in cases we do not need to 

this means we want to allow potentially bad ideas as long as they don’t break NGS when they connect to it. 

Later - 2.11 timeline and after discussions - we will decide what the new acceptable names are. 

So yea checking for extra things is not a problem and not bad - but our current aim is to capture the bare minimum cases that cause problems so we break as few as possible users.

mprimi: Well this check in `audit` doesn't break anything. 

It may be worth being a bit more pedantic ahead of the actual changes to server.

PR #1097, opened 2024-07-24
Sort subjects consistently when doing nats subscribe --report-subjects

johnweldon:
Previously, report subjects only sorted by message count, but if two subjects had the same message count, they would not be sorted consistently.

This change sorts the subjects first by count, then by subject


ripienaar:
Thank you

PR #1098, opened 2024-07-25
#181 support TLS Insecure connection - added flag --tlsinsecure

### LINKED ISSUE #181: Allow TLS verification to be skipped
Hello,

In production we're using valid TLS certificates _only_ for securing the connection, not validating the client. However in test I'm using self-signed certificates for this. There doesn't appear to be a way to disable the NATS client from attempting to verify the TLS certificate of the server. It would be nice to have this option.

Thanks

mschneider82:
This PR adds a new flag for `--tlsinsecure` or env `NATS_TLSINSECURE`
Since the server side supports it, the cli should also support it.

fixes #181 

tested in a pod:

```
nats-box:/# ./nats obj ls 
nats: error: tls: failed to verify certificate: x509: certificate is valid for ::1, 127.0.0.1, not 10.1.1.1

nats-box:/# ./nats obj ls --tlsinsecure
╭────────────────────────────────────────────────────────────────────╮
│                        Object Store Buckets                        │
├─────────┬─────────────┬─────────────────────┬────────┬─────────────┤
│ Bucket  │ Description │ Created             │ Size   │ Last Update │
│ test    │             │ 2024-07-22 08:31:17 │ 12 KiB │ 57m3s       │
╰─────────┴─────────────┴─────────────────────┴────────┴─────────────╯

```

### DIFF
--- a/cli/util.go
+++ b/cli/util.go
@@ -309,7 +309,15 @@ func natsOpts() []nats.Option {
 	}
 
 	if opts().TlsInsecure {
-		copts = append(copts, nats.Secure(&tls.Config{InsecureSkipVerify: true}))
+		insecureOption := func(o *nats.Options) error {
+			if o.TLSConfig == nil {
+				o.TLSConfig = &tls.Config{InsecureSkipVerify: true}
+			} else {
+				o.TLSConfig.InsecureSkipVerify = true
+			}
+			return nil
+		}
+		copts = append(copts, insecureOption)	
 	}
 
 	return append(copts, []nats.Option{
--- a/cli/util.go
+++ b/cli/util.go
@@ -17,6 +17,7 @@ import (
 	"bufio"
 	"bytes"
 	"context"
+	"crypto/tls"
 	"encoding/base64"
 	"encoding/json"
 	"errors"
@@ -305,6 +306,10 @@ func natsOpts() []nats.Option {
 	connectionName := strings.TrimSpace(opts().ConnectionName)
 	if len(connectionName) == 0 {
 		connectionName = "NATS CLI Version " + Version
+	}
+
+	if opts().TlsInsecure {
+		copts = append(copts, nats.Secure(&tls.Config{InsecureSkipVerify: true}))
 	}
 
 	return append(copts, []nats.Option{
--- a/cli/util.go
+++ b/cli/util.go
@@ -310,6 +310,7 @@ func natsOpts() []nats.Option {
 
 	if opts().TlsInsecure {
 		insecureOption := func(o *nats.Options) error {
+			o.Secure = true
 			if o.TLSConfig == nil {
 				o.TLSConfig = &tls.Config{InsecureSkipVerify: true}
 			} else {
--- a/cli/cli.go
+++ b/cli/cli.go
@@ -56,7 +56,7 @@ var (
 
 	// These are persisted by contexts, as properties thereof.
 	// So don't include NATS_CONTEXT in this list.
-	overrideEnvVars = []string{"NATS_URL", "NATS_USER", "NATS_PASSWORD", "NATS_CREDS", "NATS_NKEY", "NATS_CERT", "NATS_KEY", "NATS_CA", "NATS_TIMEOUT", "NATS_SOCKS_PROXY", "NATS_COLOR"}
+	overrideEnvVars = []string{"NATS_URL", "NATS_USER", "NATS_PASSWORD", "NATS_CREDS", "NATS_NKEY", "NATS_CERT", "NATS_KEY", "NATS_CA", "NATS_TIMEOUT", "NATS_SOCKS_PROXY", "NATS_COLOR", "NATS_TLSINSECURE"}
 )
 
 func registerCommand(name string, order int, c func(app commandHost)) {
--- a/nats/main.go
+++ b/nats/main.go
@@ -57,6 +57,7 @@ See 'nats cheat' for a quick cheatsheet of commands`
 	ncli.Flag("tlskey", "TLS private key").Envar("NATS_KEY").PlaceHolder("FILE").ExistingFileVar(&opts.TlsKey)
 	ncli.Flag("tlsca", "TLS certificate authority chain").Envar("NATS_CA").PlaceHolder("FILE").ExistingFileVar(&opts.TlsCA)
 	ncli.Flag("tlsfirst", "Perform TLS handshake before expecting the server greeting").BoolVar(&opts.TlsFirst)
+	ncli.Flag("tlsinsecure", "Disable TLS Certificate Verification").Envar("NATS_TLSINSECURE").BoolVar(&opts.TlsInsecure)
 	if runtime.GOOS == "windows" {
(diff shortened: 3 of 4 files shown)


cli/util.go
```diff
@@ -307,6 +308,10 @@ func natsOpts() []nats.Option {
 		connectionName = "NATS CLI Version " + Version
 	}
 
+	if opts().TlsInsecure {
+		copts = append(copts, nats.Secure(&tls.Config{InsecureSkipVerify: true}))
```
ripienaar, line 312: the problem with this approach is that it invalidates all other tls setup.  You may very well still want to send certificate and use a key but you just want to skip the CA setup.

This approach will only work with servers that dont validate your client cert.  I've previously rejected essentially this same PR because I dont think we could come up with a plan that worked better.


We would essentially have to create an entire valid tlsc with all things loaded just like the client would with just this one setting set in addition.

I think the go client should get a feature flag to enable this instead

mschneider82: I think TLS Insecure is just for usecases where the server also has insecure flag set and doesnt validate tls ceritificates. 
Instead of creating a new tls.Config{} we could use the existing tls.Config if it would be exported somewhere.

ripienaar: Yeah, I think better if we get an option into nats.go to set this flag

mschneider82: Its better to have this option with some exeptions like it doesnt support sending certificates instead of not able to connect any insecure tls server. The fix for that issue could be added later on when the go client supports it.

ripienaar: I disagree a bit I am not conformable doing essentially just YOLO TLS as the only option.

mschneider82: Ok i found another solution to this, please check my new commit

cli/util.go
```diff
@@ -307,6 +308,18 @@ func natsOpts() []nats.Option {
 		connectionName = "NATS CLI Version " + Version
 	}
 
+	if opts().TlsInsecure {
```
ripienaar, line 311: I believe custom tlsconfigs has to be set using `nats.Secure()`.

Again, this works today only due to chance and due to flow of nats.go which can change at any time.  This feature has to go into nats.go please engage there and add an option so it's done right and we know it will work and be maintained in the long term.

mschneider82: What makes me a little confusion is the Description of nats.Secure() :
```
// Secure is an Option to enable TLS secure connections that skip server verification by default.
// Pass a TLS Configuration for proper TLS.
// A TLS Configuration using InsecureSkipVerify should NOT be used in a production setting.
```

In my first PR i tried to use `nats.Secure()` to enable `InsecureSkipVerify`, this only works if i supply a `tls.Config{}` struct. So currently nats.Secure() is broken because it doesnt handle the mentioned Auth things?

ripienaar: Secure() without a tlsc sets the Secure option true

https://github.com/nats-io/nats.go/blob/c693ec3784c00cf5f9ad9c0f4010b352482ab86d/nats.go#L880

Later this is called that will create a tlsc and set all the things so just calling Secure() would be fine

https://github.com/nats-io/nats.go/blob/c693ec3784c00cf5f9ad9c0f4010b352482ab86d/nats.go#L2088

if you do want to set a tls at least also set `MinVersion: tls.VersionTLS12` here

ripienaar: But all that aside, I am still not happy supporting this option. 

Why is your TLS setup broken that you need this?

mschneider82: This is for development env, I embedded nats js server and didnt want to use plain connection, plain is way more insecure than using a self signed tls.

I have tested just calling nats.Secure() without an tls.Config and the tls failure still exists.

ripienaar: if nats.Secure() doesn't do it, then again this should be fixed in nats.go 

I strongly believe delegating tls setup to nats.go is the right thing to do - as security design, minimal versions or other constraints evolve we will forget to do same here when done in nats.go.  Relying on nats.go to do the right thing is the safest behaviour.

mschneider82: Ok i will prepare a PR for that

liaoya:
This feature is very helpful and even our production use insecure certificate.

cli/util.go
tehsphinx: Not sure I understand why this should go into `nats.go` as the only place we can't set `InsecureSkipVerify` is the NATS cli.

tehsphinx: @mschneider82: I cloned the repository, applied the changes and installed the adjusted CLI. Finally I can just connect to all our test environments and see what is going on with my JetStream subscriptions -- without having to set up each of them with the right certificates locally.

This is worth gold! 🥇

Thank you!

mschneider82: Yes the patched Version, especially with the env NATS_TLSINSECURE set is total fine. It could be so simple to implement...

ripienaar: We already support bypassing TLS - just don't set it up since disabling key aspects of it has the same outcome.

tehsphinx: > We already support bypassing TLS - just don't set it up since disabling key aspects of it has the same outcome.

Can you elaborate?

But note, that I'm not interested in bypassing TLS. I'm just interested in disabling the server certificate verification in `natsCLI`. The communication is still TLS encrypted.

ripienaar:
If setting up TLS correctly to get the benefits of TLS including man in tbe middle attack prevention is too much effort. Then don’t set up TLS and make it clear to everyone you don’t care for those features. 

Selectively disabling things like peer verification means you may as well not set up TLS.

tehsphinx:
> Selectively disabling things like peer verification means you may as well not set up TLS.

@ripienaar How is setting setting `InsecureSkipVerify` equal to disabling peer verification? The natsCLI is not a peer. It is a debugging tool. (At least for the use case I'd like to be able to use `InsecureSkipVerify` for.)

tehsphinx:
> It’s much more than a debugging tool. It’s capable of creating jetstream assets, removing them, destroying data, creating accounts and users and severely interrupting even destroying jetstream clusters.
> 
> It’s a full admin tool and equivalent to root level access for the nats server.

This doesn't answer the question.

Note, that I never said the NATS cli is merely a debug tool, but it is for the use case I need `InsecureSkipVerify`. Also note that this does not degrade the security of the cluster as I'm not changing anything on the cluster side.

Overall I see a lot of made up reasons in this MR. A lot of aggression and assumptions, too.

steveej:
i agree with @tehsphinx here.

i came across this today where i have a test deployment of our production NATS config in a container locally. it defaults to self-generated TLS certificates if it cannot get ACME verification.

please revisit this use-case (and PR).

cli/util.go
rumsrami: > > We already support bypassing TLS - just don't set it up since disabling key aspects of it has the same outcome.
> 
> Can you elaborate?
> 
> But note, that I'm not interested in bypassing TLS. I'm just interested in disabling the server certificate verification in `natsCLI`. The communication is still TLS encrypted.
> 
> Exactly what this PR does: it doesn't disable TLS. Just allow connection to a server with a certificate it can't verify against a CA.

To be honest, i do not understand. The only thing I want to do is use nats cli to connect to a server with a certificate it can't verify against a CA." This is needed only for debugging. Can the current natscli do this?

PR #1103, opened 2024-07-26
Fix whitespace check to include newlines

ReubenMathew:
#1096 only checks for ` `, while other characters (e.g., `\n` in the server name can still break the cluster.

Signed-off-by: reubenninan <reuben@nats.io>

### DIFF
--- a/internal/audit/cluster_checks.go
+++ b/internal/audit/cluster_checks.go
@@ -221,7 +221,7 @@ func makeCheckClusterHighHAAssets(haAssetsThreshold int) checkFunc {
 func checkClusterNamesForWhitespace(reader *archive.Reader, examples *ExamplesCollection) (Outcome, error) {
 
 	for _, clusterName := range reader.GetClusterNames() {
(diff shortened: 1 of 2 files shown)


ripienaar:
@ReubenMathew @mprimi lets get this one in so I can do the maintenance release asap thanks!

ReubenMathew:
> @ReubenMathew @mprimi lets get this one in so I can do the maintenance release asap thanks!

Have we decided if space and newline characters are enough to check for? Should we just be safe and use `unicode.IsSpace`?

ripienaar:
The Unicode check doesn’t solve any actual problems. So space and new line seems to address the problem.

ripienaar (commented):
LGTM

just to clarify. In meetings we agreed on 2.10 restrictions - space and you pointed out new line also so that’s good. We also discussed 2.11 likely being restricted-term. 

So Unicode checks seemed to check 2.10 but also add additional limits that won’t actually prep people for 2.11

Issue #1127, opened 2024-08-15
Improve consumer list


ripienaar:
### Proposed change

`nats s ls` will show a table by default with some basic information extracted from the stream list api response (in other words no additional API calls per stream). It also have `--names` and `--json` formats.

We should make consumer list work the same, in the table being shown:

 * name
 * description
 * created
 * ack pending
 * unprocessed
 * last delivery as a duration

Also make sure `--names` and `--json` work.

It's basically a minimal `c report` so just copy code from there and make a more minimal output

### Use case

improved ux

PR #1128, opened 2024-08-22
(#1127) Improve consumer list

ploubser:
Here we update the `consumer ls` command to behave similar to the `stream ls` command. It will now show a table including the name, description, created, ack pending, unprocessed and last delivery fields.

`consumer ls --names` and `consumer ls --json` behaviour remains unchanged.


ripienaar (approved):
LGTM

Issue #1134, opened 2024-09-01
Uiprogress runtime error during "nats bench" when the number of publishers exceeds the number of messages


giacomoquinalia:
### Observed behavior

Running a command like `nats bench foo.bar --pub 2 --msgs 1` I receive the output:

```sh
10:51:24 Starting Core NATS pub/sub benchmark [subject=foo.bar, multisubject=false, multisubjectmax=100000, msgs=1, msgsize=128 B, pubs=2, subs=0, pubsleep=0s, subsleep=0s]
10:51:24 Starting publisher, publishing 1 messages
10:51:24 Starting publisher, publishing 0 messages
panic: runtime error: index out of range [0] with length 0

goroutine 39 [running]:
github.com/gosuri/uiprogress.(*Bar).Bytes(0xc000254000)
	/home/giacomo/go/pkg/mod/github.com/gosuri/uiprogress@v0.0.1/bar.go:195 +0x40b
github.com/gosuri/uiprogress.(*Bar).String(...)
	/home/giacomo/go/pkg/mod/github.com/gosuri/uiprogress@v0.0.1/bar.go:214
github.com/gosuri/uiprogress.(*Progress).print(0xc0000a2480)
	/home/giacomo/go/pkg/mod/github.com/gosuri/uiprogress@v0.0.1/progress.go:127 +0x8f
github.com/gosuri/uiprogress.(*Progress).Listen(0xc0000a2480)
	/home/giacomo/go/pkg/mod/github.com/gosuri/uiprogress@v0.0.1/progress.go:116 +0xb3
created by github.com/gosuri/uiprogress.(*Progress).Start in goroutine 1
	/home/giacomo/go/pkg/mod/github.com/gosuri/uiprogress@v0.0.1/progress.go:134 +0x4f
```

### Expected behavior

To prevent a runtime error when the number of messages for a publisher is zero, the progress bar could be shown as already completed instead of attempting to update it. A possible solution:

https://github.com/nats-io/natscli/blob/main/cli/bench_command.go#L1718
```go
	if c.progressBar {
		barCount := numMsg
		if barCount == 0 {
			barCount = 1
		}

		progress = uiprogress.AddBar(barCount).AppendCompleted().PrependElapsed()
		progress.Width = progressWidth()

		if numMsg == 0 {
			progress.PrependFunc(func(b *uiprogress.Bar) string {
				return "Finished  "
			})
			progress.Incr()
		}
	}

	if numMsg == 0 {
[...]

PR #1138, opened 2024-09-03
Fix progress bar display issue

giacomoquinalia:
There is an issue with a Uiprogress runtime error during "nats bench" when the number of publishers exceeds the number of messages. This results in some publishers being assigned 0 messages, which is set as the total for the progress bar. However, Uiprogress does not accept zero as a valid total and throws a runtime error.

To prevent this error, the progress bar can be shown as already completed when there are no messages to publish.

Issue: https://github.com/nats-io/natscli/issues/1134


ripienaar:
Thank you

PR #1143, opened 2024-09-06
The consumer max waiting setting should not be editable

ripienaar:


ripienaar:
```
[rip@s1-lon]% nats c edit ORDERS_0 --max-waiting 1000
[lon] ? Select a Consumer C0
Differences (-old +new):
  api.ConsumerConfig{
        ... // 13 identical fields
        MaxDeliver:   -1,
        BackOff:      nil,
-       MaxWaiting:   512,
+       MaxWaiting:   1000,
        OptStartSeq:  0,
        OptStartTime: nil,
        ... // 13 identical fields
  }
[lon] ? Really edit Consumer ORDERS_0 > C0 Yes
nats: error: max waiting can not be updated (10012)
```

Issue #1145, opened 2024-09-09
Support interactive edit for consumers


ripienaar:
### Proposed change

Make it like `nats s edit -I`

### Use case

consumer editing

ripienaar:
This one is pretty broken, shows many wrong fields

```
 32 # List of durations in Go format that represents a retry time scale for NaK'd messages
 33 #
 34 #   Type: array (items: integer)
 35 maxwaiting: 512
 36 optstartseq: 0
 37 optstarttime: null
 38 ratelimit: 0
 39 replaypolicy: 0
 40 samplefrequency: ""
 41 headersonly: false
 42 maxrequestbatch: 0
 43 maxrequestexpires: 0s
 44 maxrequestmaxbytes: 0
 45 inactivethreshold: 0s
 46 replicas: 0
 47 memorystorage: false
```

Compared to https://github.com/nats-io/nats-server/blob/e485bfa816d7c62b687df23447ac28784f5b4f38/server/consumer.go#L116

ripienaar:
The issue here is that the ConsumerConfig doesn't have the yaml markup here https://github.com/nats-io/jsm.go/blob/f5cd8a22d0931e69320eb585d6265069e1326a34/api/consumers.go#L402 

As the backend for this is a yaml marshal/unmarshal without those keys it doesn't turn `MemoryStorage` into `mem_storage` instead doing `memorystorage` that isnt in the schema.

So we'll have to add the yaml tags on the main config and all related structs, code wise I think CLI is fine

PR #1151, opened 2024-09-12
(#1145) Support interactive edit for consumers

ploubser:
Here we add the --interactive flag to the consumer edit action. This behaves the same as stream interactive edit.


ripienaar:
not bad, my weird yaml marshal thing seems to be generic enough :)

PR #1152, opened 2024-09-12
Allow turn off of doReqAsync adaptive timeout

mprimi:
`doReqAsync` uses an adaptive timeout to stop listening for responses when the number of responses waited is zero.

In this PR:
 - Create special value `waitFor=-1` to wait for the full `timeout` for responses
 - Update `audit gather` wait for the full timeout

### DIFF
--- a/cli/util.go
+++ b/cli/util.go
@@ -782,6 +782,17 @@ func renderCluster(cluster *api.ClusterInfo) string {
 	return f(compact)
 }
 
+// doReqAsyncWaitFullTimeoutInterval special value to be passed as `waitFor` argument of doReqAsync to turn off
+// "adaptive" timeout and wait for the full interval
+const doReqAsyncWaitFullTimeoutInterval = -1
+
+// doReqAsync serializes and sends a request to the given subject and handles multiple responses.
+// This function uses the value from `Timeout` CLI flag as upper limit for responses gathering.
+// The value of the `waitFor` may shorten the interval during which responses are gathered:
+//
+//	waitFor < 0  : listen for responses for the full timeout interval
+//	waitFor == 0 : (adaptive timeout), after each response, wait a short amount of time for more, then stop
+//	waitFor > 0  : stops listening before the timeout if the given number of responses are received
 func doReqAsync(req any, subj string, waitFor int, nc *nats.Conn, cb func([]byte)) error {
 	jreq := []byte("{}")
 	var err error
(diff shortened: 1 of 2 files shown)


cli/util.go
```diff
@@ -852,7 +857,11 @@ func doReqAsync(req any, subj string, waitFor int, nc *nats.Conn, cb func([]byte
 			}
 		}
 
-		if finisher != nil {
+		// Set timeout for the next response to arrive.
+		// Unless:
+		//  - We are waiting for a specific number of responses (finisher is null) OR
+		//  - We were told explicitly to wait for the full interval for responses
+		if finisher != nil && waitFor != doReqAsyncWaitTimeoutInterval {
```
ripienaar, line 864: wouldn't finisher always be nil if waitFor is -1, so just checking `if finisher != nil` should be sufficient here, if not we should restructure things that the nil check is sufficient, easy to miss multiple places to update if this behaviour changes in future again.

ripienaar (changes requested):
small change only

cli/util.go
mprimi: The default timeout is 5s, so waits 5 extra seconds from the last response. 

I think "stop doing smart things and just wait a fixed amount of time" is a more useful "alternative mode" for this utility function.

ripienaar: Still convinced you should just write your own request method. They have different goals.

cli/util.go
```diff
@@ -782,6 +782,11 @@ func renderCluster(cluster *api.ClusterInfo) string {
 	return f(compact)
 }
 
+// doReqAsyncWaitFullTimeoutInterval special value to be passed as `waitFor` parameter of doReqAsync, which results in
+// the latter function waiting for the Options.Timeout duration for replies. e.g. it turns off the "adaptive"
+// timeout behavior ("wait 300ms after the previous response then stop listening")
+const doReqAsyncWaitFullTimeoutInterval = -1
```
ripienaar, line 788: imo this is harder to discover being here than just adding a comment to the `doReqAsync` function calling out this behaviour, this way it will be visible whenever someone uses the function but this constant would only be known to those going to look for it.

mprimi: Agreed, this is hard to discover and doesn't do much.

(I do like this constant for calling code, it's more readable than passing in a magic `-1`).

Updated to just document `doReqAsync`

ripienaar: nice, could have kept the constant was mainly concerned with discoverability of the constant.  Looks good now will merge. Thank you!

Issue #1140, opened 2024-09-04
Integrate internal dev tools


ripienaar:
### Proposed change

We have 2 tools developed while supporting users that we'd like to include in the cli

 * https://github.com/nats-io/nats.go/blob/stream-check/examples/consumer-check/main.go
 * https://github.com/nats-io/nats.go/blob/stream-check/examples/stream-check/main.go

Requires some thought but would be great to figure out how to best to include them

### Use case

Improve supportability

ripienaar:
cc @wallyqs who requested this and wrote these

PR #1158, opened 2024-09-24
(#1140) Add stream-check and consumer-check subcommands

ploubser:
Here we add the `stream-check` and `consumer-check` subcommands to the `server` command.

These subcommands are hidden, and will use the system credentials to iterate all servers, streams and consumers and display debugging related information to the user in a table.


ripienaar:
Thank you @ploubser

PR #1164, opened 2024-10-02
(maint) Update kv and object commands to use new jetstream api

ploubser:
Here we update both the kv and object commands to no longer use the deprecated nats.JetstreamContext and instead use jetstream.Jetstream.

We also fix some minor usability issues, like make the `revision` flag required for the `update` action, improve the error message when attempting to revert a key value in a bucket with history=1, and perform the revert with an `update` action instead of a `put`.

### DIFF
--- a/cli/kv_command.go
+++ b/cli/kv_command.go
@@ -14,6 +14,7 @@
 package cli
 
 import (
+	"context"
 	"errors"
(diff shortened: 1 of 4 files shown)


ripienaar (approved):
LGTM

PR #1167, opened 2024-10-07
Support handling panics in checks and reporting traces

ripienaar:


ripienaar:
replaced with https://github.com/nats-io/natscli/pull/1258

PR #1170, opened 2024-10-09
(maint) Make context checkFile fn expand tildes

ploubser:


cli/context_command.go
```diff
@@ -454,6 +455,14 @@ func (c *ctxCommand) showCommand(_ *fisk.ParseContext) error {
 		if strings.HasPrefix(file, "op://") {
 			return color.CyanString("1Password")
 		}
+		if file[0] == '~' {
+			usr, err := user.Current()
+			if err != nil {
```
ripienaar, line 460: return the error here so users can have some hope of figuring this out

ripienaar (changes requested):

PR #1174, opened 2024-10-22
Add support for EDITOR with args to commands

### LINKED ISSUE #1173: Support for EDITOR with args when editing context
### Proposed change

I would like to be able to use VS Code to edit my NATS context. VS Code has a `--wait` option to wait for the files to be closed before returning, but currently I get an error when using it.

```shell
$ nats context edit cluster_joe
nats: error: exec: "code --wait": executable file not found in $PATH
```

This could also be useful for any other use of `"EDITOR"`.

### Use case

I would like to edit my NATS context using VS Code with args.

### Contribution

Yes

joeriddles:
closes #1173

Add support for EDITOR with args for commands that use EDITOR, like `nats context edit`.

This allows users to use something like `code --wait` for their EDITOR.

### DIFF
--- a/cli/consumer_command.go
+++ b/cli/consumer_command.go
@@ -561,11 +561,15 @@ func (c *consumerCmd) leaderStandDownAction(_ *fisk.ParseContext) error {
 }
 
 func (c *consumerCmd) interactiveEdit(cfg api.ConsumerConfig) (*api.ConsumerConfig, error) {
-	editor := os.Getenv("EDITOR")
-	if editor == "" {
+	rawEditor := os.Getenv("EDITOR")
+	if rawEditor == "" {
 		return &api.ConsumerConfig{}, fmt.Errorf("set EDITOR environment variable to your chosen editor")
 	}
-	editor, args := splitCommand(editor)
+
+	editor, args, err := splitCommand(rawEditor)
+	if err != nil {
+		return &api.ConsumerConfig{}, fmt.Errorf("could not parse EDITOR: %v", rawEditor)
+	}
 
 	cj, err := decoratedYamlMarshal(cfg)
 	if err != nil {
--- a/cli/consumer_command.go
+++ b/cli/consumer_command.go
@@ -22,7 +22,6 @@ import (
 	"math"
 	"math/rand"
 	"os"
-	"os/exec"
 	"os/signal"
 	"regexp"
 	"sort"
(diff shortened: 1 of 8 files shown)


cli/util.go
```diff
@@ -254,6 +254,14 @@ func splitString(s string) []string {
 	})
 }
 
+// Split the string into a command and its arguments.
+func splitCommand(s string) (string, []string) {
+	cmdAndArgs := strings.Fields(s)
```
ripienaar, line 259: Better use `shlex.Split()` this is shell aware

joeriddles: Updated to use `shlex.Split()`: https://github.com/nats-io/natscli/pull/1174/commits/48fbe2963604a9af4810410e04a683c03a93b490.

@ripienaar do you think it's better to return `(string, []string, error)` or return a result struct + error?

cli/consumer_command.go
```diff
@@ -561,11 +561,16 @@ func (c *consumerCmd) leaderStandDownAction(_ *fisk.ParseContext) error {
 }
 
 func (c *consumerCmd) interactiveEdit(cfg api.ConsumerConfig) (*api.ConsumerConfig, error) {
-	editor := os.Getenv("EDITOR")
-	if editor == "" {
+	rawEditor := os.Getenv("EDITOR")
```
ripienaar, line 564: this seems the same everywhere, we can probably make a helper in `internal/util` something like `EditFile(f string)` and use that in all the following cases, that would be a nice improvement.

But if its too much for you to do that this PR is fine also let us know

joeriddles: Can do! Will update the PR shortly.

joeriddles: @ripienaar done, and added new tests for it using `echo` as the EDITOR command: https://github.com/nats-io/natscli/pull/1174/commits/0d162c2a8ab1db53711aa34a3cf097f3f164c4e5

ripienaar:
Looking good, I am slowly moving these utils into https://github.com/nats-io/natscli/blob/main/internal/util/util.go and its test to not pollute this package so much, be great to see these helpers moved there then we're good.

Sorry wasn't clear in earlier review about internal

ripienaar:
Some typos also

joeriddles:
Ah makes sense, I will move the edit util func over

PR #1176, opened 2024-11-01
Update the readme's nats bench examples to use the new nats bench commands.

jnmoyne:
Noticed that had not been updated with the new nats bench


ripienaar:
Looks good but lets park this till just before next release so info currently shown stays relevant.

Just waiting really for the versioning work to settle in the server will do a release then

PR #1177, opened 2024-11-01
Bench command in multisubject mode subject format improvement

jnmoyne:
When using nats bench in multisubject mode to generate messages in order to evaluate the amount of memory used by the indexes according to the number of subjects being stored in the stream it is more accurate if all the subjects have the same size. This achieves that by using padding 0 up to the length of the max number of subjects when creating the subjects.


ripienaar (approved):
LGTM

ripienaar:
Please sign your commit @jnmoyne

jnmoyne:
Apologies: they were signed but with my new GPG key (old one expired) and I just had not uploaded that new key to GH yet, done now.

PR #1180, opened 2024-11-14
Fork and update guptarohit/asciigraph

ploubser:
Here we fork guptarohit/asciigraph and add it to our internal packages.

We have added the ability to specify a `ValueFormatter` option for a graph, allowing us to prettify the lables shown on the y-axis of a graph.

With that we have added the ability to specify `AlwaysY`, which will force the graph to draw the y-axis even in cases where the input data is 0, allowing us to draw multiple graphs in the terminal with consistent size and clariy.


ripienaar (changes requested):
small comment otherwise good stuff

ripienaar (approved):
LGTM

PR #1191, opened 2024-12-09
goreleaser: switch to v2 config & GH Action

philpennock:
Handle deprecated configuration settings, updating per guidance in <https://goreleaser.com/deprecations/>.

Add an internal `version:` field in the GoReleaser YAML, for v2.

Switch the GitHub Action to @v6 which will use goreleaser v2, and make the version explicitly `"~> v2"`.  Before, it said "latest" but that was implicitly bound by the action to be v1, leading to a little head-scratching.


ripienaar (approved):
LGTM

PR #1192, opened 2024-12-09
avoid int overflow compile errors for 32-bit target

philpennock:


philpennock:
Oops, this includes the #1191 commit and should not.  Failing to recall how to avoid that, without a lot more branch work.  Probably easiest to just land 1191 first and then rebase this.  Sorry.

ripienaar:
Thanks

PR #1193, opened 2024-12-10
Bench command retry publication when pub ack future is an error

jnmoyne:
With the new changes to the max IPQ size on server v2.11 it is really easy to have some of the JS publish requests dropped when running nats bench using more than just a handfull of client connections.

This change just logs the error from a pub ack future and retries the publication instead of fatally stopping when this happens.


ripienaar (approved):
LGTM

PR #1196, opened 2024-12-11
Clarify duration format

alexbozhenko:
For someone who is not familiar with Golang's duration type(https://pkg.go.dev/time#ParseDuration), this clarification may be useful.

Before:
```
[alex@omenlaptop natscli]$ nats bench pub -h
usage: nats bench pub [<flags>] <subject>

Publish Core NATS messages

Args:
  <subject>  Subject to use for the benchmark

Flags:
  --sleep=DURATION          Sleep for the specified interval between publications
  --clients=1               Number of concurrent clients
```

After:
```
[alex@omenlaptop natscli]$ go run ./nats bench pub -h
usage: nats bench pub [<flags>] <subject>

Publish Core NATS messages

Args:
  <subject>  Subject to use for the benchmark

Flags:
  --sleep=DURATION          Sleep for the specified interval between publications, such as 300ms or 1m3s. Valid time units are "ns", "us", "ms",
                            "s", "m", "h"
  --clients=1               Number of concurrent clients
```

Should we add the same to other places where  
[DurationVar](https://github.com/search?q=%2F%28%3F-i%29DurationVar%2F+repo%3Anats-io%2Fnatscli+lang%3AGo+&type=code) is used?

### DIFF
--- a/cli/bench_command.go
+++ b/cli/bench_command.go
@@ -112,7 +112,7 @@ func configureBenchCommand(app commandHost) {
 
 	addJSCommonFlags := func(f *fisk.CmdClause) {
 		f.Flag("stream", "The name of the stream to create or use").Default(benchDefaultStreamName).StringVar(&c.streamOrBucketName)
-		f.Flag("sleep", "Sleep for the specified interval between publications").Default("0s").PlaceHolder("DURATION").DurationVar(&c.sleep)
+		f.Flag("sleep", "Sleep for the specified interval between publications.").Default("0s").PlaceHolder("DURATION").DurationVar(&c.sleep)
 		f.Flag("purge", "Purge the stream before running").UnNegatableBoolVar(&c.purge)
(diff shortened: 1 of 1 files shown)


ripienaar:
There are 54 flags or arguments that take durations.  Many more prompts and other areas where durations are accepted.

We cannot add this to each one, this is better handled as a section in the README.

ripienaar:
In addition to standard go duration strings we also accept:

```
// * "w", "W" - a week based on 7 days of exactly 24 hours
// * "d", "D" - a day based on 24 hours
// * "M" - a month made of 30 days of 24 hours
// * "y", "Y" - a year made of 365 days of 24 hours each
```

alexbozhenko:
@ripienaar 
So where exactly should we add it?

ripienaar:
A section in the readme, take a look where it would fit. Could also add it to the website docs

PR #1199, opened 2024-12-13
Add a --match-replies-timeout to print messages without a reply.

apollo13:


cli/sub_command.go
```diff
@@ -110,6 +111,7 @@ func configureSubCommand(app commandHost) {
 	// To be done - check for streams with WorkQueue, then prompt with or allow with override --force=WorkQueueDelete
 	// act.Flag("ackPolicy", "Acknowledgment policy (none, all, explicit) (requires JetStream)").Default("none").EnumVar(&c.ackPolicy, "none", "all", "explicit")
 	act.Flag("match-replies", "Match replies to requests").UnNegatableBoolVar(&c.match)
+	act.Flag("match-replies-timeout", "Timeout after which messages without a reply are printed").PlaceHolder("TIMEOUT").DurationVar(&c.matchTimeout)
```
apollo13, line 114: I'd love to set this to a (low?) default.

ripienaar: Rather just use the overall timeout set with --timeout, dont think its needed add more ways to set timeouts.  That one already have a default.

Still, I am not sure if I want to add this, will give it some thought.

apollo13: Let me know if you want it, then I'll adjust the PR. Otherwise feel free to close it.

ripienaar: Sorry fell below the radar during the holidays.

I think I'll pass on this for the moment, things get a bit complicated and hard to understand with every new flag and behaviour so want to be quite careful about it.

If others report similar concerns we can revisit, sorry about that.

PR #1221, opened 2024-12-31
Allow updating KV bucket settings from add command

kroepke:
add --[no]-update flag to add KV command

if set, the update flag will use CreateOrUpdateKeyValue instead of CreateKeyValue to allow changing settings of the bucket

Not passing the parameter retains the original behavior

### DIFF
--- a/cli/kv_command.go
+++ b/cli/kv_command.go
@@ -65,6 +65,7 @@ type kvCommand struct {
 	mirrorDomain          string
 	sources               []string
 	compression           bool
+	update                bool
 }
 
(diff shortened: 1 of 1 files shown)


ripienaar:
Thanks for this, I think I'd rather model this around how `nats stream` works with a `add` and `edit` commands separate and edit only showing flags that's changeable

ripienaar:
Supporting updates to bucket config has been on my todo list but never got to it, so would be very grateful to see a contribution there!

kroepke:
Closing this in favor of #1240

PR #1227, opened 2025-01-02
Add man page and zsh completion

GaMoCh:
Please, update the [fisk](github.com/choria-io/fisk) dependency version after https://github.com/choria-io/fisk/pull/54 merge.

[36f3fec](https://github.com/nats-io/natscli/commit/36f3fec3581a114f93fad3adb2f12b768d80f699) uses [zsh.complete.nats](https://github.com/ConnectEverything/client-tools/blob/main/zsh.complete.nats), but I consider that `generate_completions_from_executable` is more suitable for maintenance.

### DIFF
--- a/.goreleaser.yml
+++ b/.goreleaser.yml
@@ -66,8 +66,17 @@ brews:
     skip_upload: auto
     test: |
       system "#{bin}/nats --version"
+    custom_block: |
+      resource "zsh_completion" do
+        url "https://get-nats.io/zsh.complete.nats"
+        sha256 "2b0832b26b1644ff01a293dc1b91320fc24f818ad8e89dd416f9c83f23b4df59"
+      end
     install: |
       bin.install "nats"
+      (man1/"nats.1").write Utils.safe_popen_read(bin/"nats", "--help-man")
+      resource("zsh_completion").stage do
+        zsh_completion.install "zsh.complete.nats" => "_nats"
+      end
 
 nfpms:
   - file_name_template: 'nats-{{.Version}}-{{.Arch}}{{if .Arm}}{{.Arm}}{{end}}'
--- a/.goreleaser.yml
+++ b/.goreleaser.yml
@@ -66,17 +66,10 @@ brews:
     skip_upload: auto
     test: |
       system "#{bin}/nats --version"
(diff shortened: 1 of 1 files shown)


.goreleaser.yml
```diff
@@ -68,6 +68,8 @@ brews:
       system "#{bin}/nats --version"
     install: |
       bin.install "nats"
+      (man1/"nats.1").write Utils.safe_popen_read(bin/"nats", "--help-man")
```
ripienaar, line 71: while I like the completion thing I think the man page isnt really useful as auto generated.

So maybe drop the man page and pull in the Fisk main dependency and we can merge it, I'll do a Fisk release before natscli

GaMoCh: > I'll do a Fisk release before natscli

Thank you.

> So maybe drop the man page and pull in the Fisk main dependency and we can merge it

I can do this in this PR after the Fisk release.

> I think the man page isnt really useful as auto generated.

The man page may be useful and has already been generated. Is it really better for you to discard it?

man_page

ripienaar: Yeah I mean it’s just the help output. That’s accessible easily in the command itself. And I did work to improve how help works over kingpin so it’s super accessible when using the command. 

A man page would go into details about usage patterns and explain CLI specific concepts like contexts. 

Just showing all the help as man page is just a cop out and low effort poor outcome.

GaMoCh: Man page removed: [fd30578](https://github.com/nats-io/natscli/pull/1227/commits/fd30578fc8fe44efaf31823622b154aec12ebf06)

ripienaar:
Thank you @GaMoCh

.goreleaser.yml
alexbozhenko: @ripienaar 
For someone who is not ultimately familiar with the plethora of commands the tool has to offer, having help output from all the commands on a single page has the benefit that it is easily searchable.
Do you think we can commit the output of `--help-long` somewhere?
Maybe similar to how nsc publishes it with gh pages: https://nats-io.github.io/nsc/
(but using a single-page output)
It would also probably help search engines(and LLMs)

PR #1240, opened 2025-01-17
Add kv edit command

kroepke:
As discussed in #1221 this change adds an `edit` command to update KV bucket's config.

Only the updatable properties are exposed as flags.


ripienaar (approved):
LGTM

kroepke:
@ripienaar this should do the trick, please let me know if I messed this up

kroepke:
@ripienaar ok, finally

ripienaar:
Thank you @kroepke

Issue #1245, opened 2025-01-21
nats sub --dump output does not respect --translate


Zambito1:
### Observed behavior

When using `nats sub --dump` with `--translate`, the `--translate` flag is completely ignored. The `Data` field in the dump is a base64 encoding of the raw input message.

### Expected behavior

The `Data` field should be a base64 encoding of the output of passing the raw message to the command specified by the `--translate` flag.

### Server and client version

Server: `2.10.24`

Client: `v0.1.2-0.20250121111352-e105ef6dd240`

### Host environment

Running the official Docker containers on a 64 bit GNU/Linux machine.

### Steps to reproduce

* Create a stream
* Publish messages to the stream
* Dump the messages to a directory using `nats sub --stream <stream_name> --all --dump=/some/dir --translate='wc -c'`
* Observe that the `.Data` field in each of the output files in `/some/dir` is the base64 encoding of the original message, rather than the base64 encoding of the output of `wc -c` applied to each message.

ripienaar:
I think this is as desired, dump should dump the raw message as is imo, if anything we should just error if both are set

Zambito1:
The reason I am looking for this feature is because I am using protobuf for my messages, but I would like to be able to use `--translate` to dump my messages in a more portable format (JSON). This would let me easily do things like using DuckDB for analytical queries on a dumped stream (which has a `from_base64` function and handles JSON much better than it handles protobuf).

I could write a tool to do this, but since I already have a tool to translate my messages to JSON that I use with `nats sub --translate`, it seems like it would make sense to be able to combine this with `--dump`. This seems like a use case where others would find useful to reuse `--translate` tools with `--dump` too.

ripienaar:
Probably not difficult to make the command you call with translate write to a file also?

Let me think about it

Zambito1:
Using a side effecting program with `--translate` doesn't seem right to me, looking at the [documentation](https://github.com/nats-io/natscli?tab=readme-ov-file#translating-message-data-using-a-converter-command) for it. Technically it should be fine, but `--translate` being applied before `--dump` just seems like the least surprising behavior to me.

In my head, `--translate` is applied, and then the result of that is sent to the output, whether that output is the formatted stdout of `nats sub`, the `--raw` stdout of `nats sub`, or the `--dump` output.

ripienaar:
If you can build main you can test if --translate now works with --dump, thanks @Zambito1

Zambito1:
Thanks, this works for me!

PR #1259, opened 2025-01-30
(#1245) Allow --translate flag to work with --dump flag

ploubser:


cli/sub_command.go
```diff
@@ -761,6 +761,15 @@ func (c *subCmd) dumpMsg(msg *nats.Msg, stdout bool, filepath string, ctr uint)
 	serMsg.Data = msg.Data
 	serMsg.Reply = msg.Reply
 
+	if c.translate != "" {
+		data, err := filterDataThroughCmd(msg.Data, c.translate, "", "")
+		if err != nil {
+			fmt.Printf("%q\nError while translating msg body: %s\n\n", data, err.Error())
```
ripienaar, line 767: below we use log.Printf() for logging errors, should probably do same here

also needs rebase

PR #1265, opened 2025-02-04
(#12633) Fix edit when not passing --compression flag

ploubser:
In the past when a user ran `stream edit` without the `--compression` flag, the empty flag would get parsed as the compression type `none`. This means that if compression was set to `s2`, every edit without the flag would look like there was a change to the stream's compression.

In this commit we remove the default value of the compression flag, and only change the configuration if the value has been explicitly set.

### DIFF
--- a/cli/stream_command.go
+++ b/cli/stream_command.go
@@ -125,6 +125,7 @@ type streamCmd struct {
 	metadata               map[string]string
 	metadataIsSet          bool
 	compression            string
+	compressionSet         bool
 	firstSeq               uint64
 	limitInactiveThreshold time.Duration
 	limitMaxAckPending     int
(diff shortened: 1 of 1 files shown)


cli/stream_command.go
```diff
@@ -192,7 +192,7 @@ func configureStreamCommand(app commandHost) {
 		if !edit {
 			f.Flag("storage", "Storage backend to use (file, memory)").EnumVar(&c.storage, "file", "f", "memory", "m")
 		}
-		f.Flag("compression", "Compression algorithm to use (file storage only)").Default("none").EnumVar(&c.compression, "none", "s2")
+		f.Flag("compression", "Compression algorithm to use (file storage only). Defaults to 'none'").EnumVar(&c.compression, "none", "s2")
```
ripienaar, line 195: you could instead use `IsSetByUser()` to determine if the value was given on the cli or should be left alone, thats the pattern used here in other places so prefer we do that for consistency.

ploubser: 👍

Issue #1268, opened 2025-02-11
<nats auth> "mappings" section in accounts cannot be configured


roeschter:
### Proposed change

On occasion please add  'nats auth account mappings ..' sub command to allow adding and removing subject mappings.
Should be similar in syntax to imports/exports sub command.

### Use case

Allow configuration of subject mappings.

ripienaar:
Thanks @roeschter we added this, do you want to give it a spin?

PR #1285, opened 2025-02-26
(#1268) Add mappings action to auth command

ploubser:
mappings has `add`, `rm`, `list` and `info` commands.

Add can take a --config flag pointing at a valid jwt, which it will then parse and extract the mappings from it.

### DIFF
--- a/cli/auth_account_command.go
+++ b/cli/auth_account_command.go
@@ -24,6 +24,7 @@ import (
 
 	au "github.com/nats-io/natscli/internal/auth"
 	iu "github.com/nats-io/natscli/internal/util"
+	"gopkg.in/yaml.v3"
 
 	"github.com/AlecAivazis/survey/v2"
 	"github.com/choria-io/fisk"
@@ -102,6 +103,11 @@ type authAccountCommand struct {
 	tags                    []string
 	rmTags                  []string
 	signingKey              string
+	mapSource               string
+	mapTarget               string
+	mapWeight               uint
+	mapCluster              string
+	inputFile               string
 }
 
 func configureAuthAccountCommand(auth commandHost) {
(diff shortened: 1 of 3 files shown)


cli/auth_account_command.go
```diff
@@ -280,6 +285,30 @@ func configureAuthAccountCommand(auth commandHost) {
...
+
+	mappings := acct.Command("mappings", "Manage account level subject mapping and partitioning").Alias("m")
+
+	mappingsaAdd := mappings.Command("add", "Add a new mapping").Alias("new").Alias("a").Action(c.mappingAddAction)
+	mappingsaAdd.Arg("account", "Account to create the mappings on").StringVar(&c.accountName)
+	mappingsaAdd.Arg("source", "The source subject of the mapping").StringVar(&c.mapSource)
+	mappingsaAdd.Arg("target", "The target subject of the mapping").StringVar(&c.mapTarget)
+	mappingsaAdd.Arg("weight", "The weight (%) of the mappingmapping").Default("100").UintVar(&c.mapWeight)
+	mappingsaAdd.Flag("operator", "Operator to act on").StringVar(&c.operatorName)
+	mappingsaAdd.Flag("config", "JWT file to read configuration from").ExistingFileVar(&c.inputFile)
```
ripienaar, line 297: hmm, not sure we can really say read from a JWT, the idea with loading is that users can easily write a mapping in a file and load just that mapping, so we can kind of need to just load the `jwt.Mapping` from a file in json/yaml format

So I dont think for that use case users really would have JWTs around, its a convenient way to avoid all the flags and stuff

ripienaar (changes requested):
looks great other than that thing about the file, probably I wasn't clear about what I wanted :)

ripienaar (approved):
LGTM

PR #1295, opened 2025-03-11
Add payload and header flags for bench command

jnmoyne:
for: `bench pub`, `bench js pub` and `bench service request`: Adds the ability to read the message payload from a file with the `--payload` flag and to set headers on the message with `--header`.

### DIFF
--- a/cli/bench_command.go
+++ b/cli/bench_command.go
@@ -66,6 +66,8 @@ type benchCmd struct {
 	deDuplicationWindow  time.Duration
 	ack                  bool
 	randomizeGets        int
+	payloadFilename      string
+	hdrs                 []string
 }
 
 const (
@@ -107,6 +109,8 @@ func configureBenchCommand(app commandHost) {
 	addPubFlags := func(f *fisk.CmdClause) {
 		f.Flag("multisubject", "Multi-subject mode, each message is published on a subject that includes the publisher's message sequence number as a token").UnNegatableBoolVar(&c.multiSubject)
 		f.Flag("multisubjectmax", "The maximum number of subjects to use in multi-subject mode (0 means no max)").Default("100000").IntVar(&c.multiSubjectMax)
+		f.Flag("payload", "File containing a message payload to send").ExistingFileVar(&c.payloadFilename)
+		f.Flag("header", "Adds headers to the message using K:V format").Short('H').StringsVar(&c.hdrs)
 	}
 
 	addJSCommonFlags := func(f *fisk.CmdClause) {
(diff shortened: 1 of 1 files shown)


cli/bench_command.go
```diff
@@ -1534,9 +1541,43 @@ func (c *benchCmd) oldjsPullAction(_ *fisk.ParseContext) error {
...
+		if err != nil {
+			return nil, fmt.Errorf("opening the payload file: %w", err)
+		}
+		defer file.Close()
+
+		// read the file into the message payload
+		buffer, err := io.ReadAll(file)
+		if err != nil {
+			return nil, fmt.Errorf("reading the payload file: %w", err)
+		}
```
ripienaar, line 1557: this can all be `return os.ReadFile(f)`

ripienaar (changes requested):
looks good, just the one simplification

cli/bench_command.go
```diff
@@ -107,6 +109,8 @@ func configureBenchCommand(app commandHost) {
 	addPubFlags := func(f *fisk.CmdClause) {
 		f.Flag("multisubject", "Multi-subject mode, each message is published on a subject that includes the publisher's message sequence number as a token").UnNegatableBoolVar(&c.multiSubject)
 		f.Flag("multisubjectmax", "The maximum number of subjects to use in multi-subject mode (0 means no max)").Default("100000").IntVar(&c.multiSubjectMax)
+		f.Flag("payload", "File containing a message payload to send").StringVar(&c.payloadFilename)
```
ripienaar, line 112: using `ExistingFileVar()` instead of `StringVar()` and Fisk will handle error handling for non existing files for you

cli/bench_command.go
```diff
@@ -163,6 +167,8 @@ func configureBenchCommand(app commandHost) {
 	request := microService.Command("request", "Send a request and wait for its reply").Action(c.requestAction)
 	request.Help("Send a request and wait for a reply")
 	request.Arg("subject", "Subject to use for the benchmark").Required().StringVar(&c.subject)
+	request.Flag("payload", "File containing the payload to send").StringVar(&c.payloadFilename)
```
ripienaar, line 170: as above

ripienaar (changes requested):
Looks great, just the small request, sorry I missed that first time round

cli/bench_command.go
jnmoyne: Didn't know about ExistingFile/Dir, neat!

ripienaar (approved):
LGTM

PR #1313, opened 2025-03-20
nats server check credential --context-credential

philpennock:
A common use-case for checking credentials is "check my current credentials". Rather than force users to manually grab the path from their context, provide a flag `--context-credential` to just use the credential from that.

Changes in existing behavior: a different error flow for missing `--credential` with this new flag is not given.

I couldn't see a way in the fisk library to have a mandatory mutually-exclusive flag group, so coded up the checks manually when dropping `Required()` from `--credential`.


philpennock:
(force-push was an amend to fix a typo in the commit message, no change to tree)

PR #1322, opened 2025-03-28
Add --watch flag to server report action

ploubser:
Here we add a --watch x flag to the server report action. When called with this flag execution will not terminate and instead refresh the results every x seconds.

### DIFF
--- a/cli/server_report_command.go
+++ b/cli/server_report_command.go
@@ -18,7 +18,9 @@ import (
 	"errors"
 	"fmt"
 	"os"
+	"os/signal"
 	"sort"
+	"syscall"
 	"time"
 
 	iu "github.com/nats-io/natscli/internal/util"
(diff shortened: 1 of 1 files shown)


cli/server_report_command.go
```diff
@@ -122,30 +126,55 @@ func configureServerReportCommand(srv *fisk.CmdClause) {
...
 
-	routes := report.Command("routes", "Report on Route (Cluster) connections").Alias("route").Action(c.reportRoute)
+	routes := report.Command("routes", "Report on Route (Cluster) connections").Alias("route").Action(c.withWatcher(c.reportRoute))
 	routes.Arg("limit", "Limit the responses to a certain amount of servers").IntVar(&c.waitFor)
 	routes.Flag("sort", "Sort by a specific property (server,cluster,name,account,subs,in-bytes,out-bytes)").EnumVar(&c.sort, "server", "cluster", "name", "account", "subs", "in-bytes", "out-bytes")
 	addFilterOpts(routes)
 }
 
+func (c *SrvReportCmd) withWatcher(fn func(*fisk.ParseContext) error) func(*fisk.ParseContext) error {
+	return func(fctx *fisk.ParseContext) error {
```
ripienaar, line 154: we have to do a bit more here, should set up the `nc` here and pass it into each worker because we need it to not reconnect between each call which is slow and expensive

PR #1340, opened 2025-04-15
Fix account report connections

maxarndt:
#1322 broke the `nats account report connections` command. This should fix the regression.


ripienaar:
Thank you. Work has already started to expand the test coverage here. Appreciate the help.

PR #1348, opened 2025-04-21
UPDATED relevant help and logs about stream view page limit

### LINKED ISSUE #1283: nats stream view
### Proposed change

It would be nice if when having 
` nats stream view <STREAM> 100`
it would print a warning that the page size exceeds the max and it is using the max.

### Use case

Inspecting output of  nats stream view <STREAM>

### Contribution

https://github.com/nats-io/natscli/blob/519fb95a9ad03f464fac86434845c0fab390cfbf/cli/stream_command.go#L1082-L1084

Imanghvs:
Closes #1283 

Shows the 25 page size limit in helps, and also logs the limit when it is exceeded (both ideas are discussed in the issue)


ripienaar:
Thanks

Issue #1345, opened 2025-04-17
Support sending on newline for `nats pub`


bplaxco:
### Proposed change

Provide a CLI flag for `nats pub <topic>` to tell it to send a message on every newline.

Example of sending on newline:

```sh
cat requests.jsonl | nats pub request --send-on-newline
```

Or if instead you wanted it more flexible for future use cases:

```sh
# What I would like it to do
cat requests.jsonl | nats pub request --send-on=newline
# The default and what it does today
cat requests.jsonl | nats pub request --send-on=eof
```

-----

Note: If this feature already exists or is possible somewhere, my apologies for the request. The idea is based off another's feedback + me skimming the docs and help options to see if something was missed. I didn't see an option there.

### Use case

I have a tool that can listen on stdin and publish on stdout in JSONL format.

It would be nice to be able to do:

```sh
nats sub request | my-command | nats pub response --send-on=newline
```

Or even for simple transforms like:

```sh
nats sub raw | jq -cr 'some-transform' | nats pub formatted --send-on=newline
```

### Contribution

Yep! If the proposal is accepted, I'd be up for me (or someone on my team) submitting a PR with the things needed to get the feature added ^_^

ripienaar:
I think it's a good idea, would be happy to review.  Would need to be careful and support it for both JetStream and non JetStream publishing.

bplaxco:
>  Would need to be careful and support it for both JetStream and non JetStream publishing.

Okay, thanks for that note. ^_^

alexbozhenko:
Does it have to be part of the natscli?

I feel like one can achieve that with [xargs](https://www.man7.org/linux/man-pages/man1/xargs.1.html)?
```
cat /tmp/jsonl.txt | xargs -I % nats pub subject "%"
```

It has this flag:
```
       --delimiter=delim, -d delim
              Input items are terminated by the specified character. 
```

thewizzy:
> Does it have to be part of the natscli?
> 
> I feel like one can achieve that with [xargs](https://www.man7.org/linux/man-pages/man1/xargs.1.html)?
> 
> ```
> cat /tmp/jsonl.txt | xargs -I % nats pub subject "%"
> ```
> 
> It has this flag:
> 
> ```
>        --delimiter=delim, -d delim
>               Input items are terminated by the specified character. ...
> ```
> 
> That would allow split input on any char.

It is correct, this could be achieved with xargs, though it could catch delimiters (like newlines \n) in the body. 

[...]

PR #1356, opened 2025-04-26
Support sending on newline for `nats pub`

thewizzy:
Implements #1345 

The general purpose is to allow programs to pipe into natscli and have it send a message for each newline. 

* Implements `--send-on=newline`, also supports `eof`` which is considered the default behavior. 
* Implements SendOn as a type so that comparisons were quicker.
* Implements a signal.NotifyContext to terminate the program more sanely. Similar pattern to other places.
* Implemented `--quiet`/`-q` for pub to minimize non error (INFO) related logging. Not a show stopped but difficult to implement this globally without overhauling logging. 

Tested for both normal and JetStream publishing.


cli/pub_command.go
```diff
@@ -80,6 +118,11 @@ Available template functions are:
 	pub.Flag("sleep", "When publishing multiple messages, sleep between publishes").DurationVar(&c.sleep)
 	pub.Flag("force-stdin", "Force reading from stdin").UnNegatableBoolVar(&c.forceStdin)
 	pub.Flag("jetstream", "Publish messages to jetstream").Short('J').UnNegatableBoolVar(&c.jetstream)
+	pub.Flag("send-on", "When to send data from stdin: 'eof' (default) or 'newline'").Default("eof").Action(func(v *fisk.ParseContext) error {
```
ripienaar, line 121: Think we dont need the whole type and String(), Set() etc - elsewhere we use a `string` and then the flag `EnumVar()` feature like this:

https://github.com/nats-io/natscli/blob/456256a702ede68bb24aef62fdd4cae1136cfc3f/cli/stream_command.go#L191

That works fine and is much easier.

cli/pub_command.go
```diff
@@ -301,88 +344,137 @@ func (c *pubCmd) doJetstream(nc *nats.Conn, progress *progress.Tracker) error {
...
 }
 
+// readLine reads a full line from a bufio.Reader regardless of buffer size,
+// this is important for reading lines longer than the bufio.Reader's default
+// buffer size and nicer than wholesale enlarging the buffer.
+func readLine(reader *bufio.Reader) (string, error) {
+	var buf bytes.Buffer
+
+	for {
+		line, isPrefix, err := reader.ReadLine()
```
ripienaar, line 354: Would a `buffio.Scanner()` be easier here?

cli/pub_command.go
```diff
@@ -301,88 +344,137 @@ func (c *pubCmd) doJetstream(nc *nats.Conn, progress *progress.Tracker) error {
...
+		buf.Write(line)
+
+		if err != nil || !isPrefix {
```
ripienaar, line 357: If the readline above fails would we want to Write() it into buff? Seems weird to have error handling after the write

ripienaar (changes requested):
Some basic comments - will do a proper look at the logic next week

cli/pub_command.go
thewizzy: I initially had it this way. With string comparisons in a few places. 
Realistically, whilst the current way is a few orders of magnitude quicker, its probably insignificant in the overall scheme of each message. If strings are your preference, happy to change back.

ripienaar: Yeah, it would be good to stick to established patterns I think

cli/pub_command.go
thewizzy: Agree it does seem odd. 

If `bufio.ReadlLine()` fails then `line` is `nil` and `buf.Write(nil)` which has a length of 0 and nothing happens. It does make the return path simpler.

I can rewrite it as the following or add a comment to explain the oddity?
```
	for {
		line, isPrefix, err := reader.ReadLine()

		if err != nil {
			return buf.String(), err
		}

		buf.Write(line)
		if !isPrefix {
			return buf.String(), err
		}
	}
```

cli/pub_command.go
thewizzy: If you have a line longer than 64KB you will get a `token too long` error and it will truncate the input (not great with a default max message of 1MB). We have run into this in a few places and this has been our go to solution. It is generally preferred over defaulting to a much larger buffer as this would be allocated for every case and most people will probably not hit this. 

`io.ReadAll()` does not experience this issue as it keeps reading and allocating memory.

cli/pub_command.go
```diff
@@ -33,38 +33,11 @@ import (
 	terminal "golang.org/x/term"
 )
 
-type SendOn int
-
 const (
-	SendOnEOF SendOn = iota
-	SendOnNewline
+	SendOnEOF     = "eof"
```
ripienaar, line 37: these dont need to be exported

cli/pub_command.go
ripienaar: yeah that seems a bit less surprising.  No blank line above error checking.

cli/pub_command.go
```diff
@@ -301,88 +315,137 @@ func (c *pubCmd) doJetstream(nc *nats.Conn, progress *progress.Tracker) error {
...
-			return err
-		}
-		c.body = string(body)
 	}
+	for {
+		select {
+		case <-ctx.Done():
+			return fmt.Errorf("interrupted")
+		default:
+			{
```
ripienaar, line 351: not sure adding a block here adds any value other than make the code more annoying

ripienaar (changes requested):
There seems to be quite a significant issue here, my understanding is that this should works:

```
$ echo "test\npub\ninput"|bin/nats pub --send-on newline x
```

Should send 3 messages and terminate.  Instead it sends just `test` followed by infinity of empty body messages in a tight loop

```
$ echo "test\npub\ninput"|nats pub --send-on newline x
$ nats sub x --count 5
08:28:33 Subscribing on x
[#1] Received on "x"
test


[#2] Received on "x"
nil body


[#3] Received on "x"
nil body


[#4] Received on "x"
nil body


[#5] Received on "x"
nil body
```

ripienaar:
Getting nearer, the above test command never exits once EOF is reached. May be worth adding some tests into `nats/test` directory there are some examples.  You can focus just on this behaviour we will circle back and flesh it out with more tests

ripienaar:
Also just doing `nats pub --send-on newline x x` will again forever publish the same bytes

thewizzy:
Sorry if my push pulled you back in early. The current commit handles the cases you have shown. However, it will send an empty message on a line with no content, it will also send an empty message on SIGINT. Considering whether it is worthwhile having the cancellation since it cannot interrupt a readline.

I appreciate you looking into this. 

> There seems to be quite a significant issue here, my understanding is that this should works:
> 
> ```
> $ echo "test\npub\ninput"|bin/nats pub --send-on newline x
> ```
Yes, I was not handling EOF properly and was creating a new reader each time(losing buffer). `bufio.ReadLine` does not evaluate literal `\n` in a string as a newline, only "real" newlines - and having it evaluate newlines in a string would likely introduce issues. 

The following will work as `-e` interprets the escapes creating "real" newlines.
```
$ echo -e "test\npub\ninput"|bin/nats pub --send-on newline x
```

ripienaar:
Looks good - we must have adjusted main since your fork and probably moved some tests around, if you fix those I think we can mark is as ready

thewizzy:
> Looks good - we must have adjusted main since your fork and probably moved some tests around, if you fix those I think we can mark is as ready

Thank you for looking into it. What are your thoughts on removing the signal handling? Because of the way `bufio.ReadLine` works we cannot interrupt it properly until we get a newline - could send an empty message. Not fussed either way.

Also, I have been testing on our project and we get a few errors like `template: body:1: function "lookup" not defined`. The issue being we will sometimes have `{{ template }}` in the messages. It would be good to implement another option `--no-templates` or similar. Would you prefer a separate PR for this?

Thank you again for your efforts reviewing my PR and feedback.

ripienaar:
Yeah I think a flag like that will be good, the way we do them is like this:

https://github.com/nats-io/natscli/blob/456256a702ede68bb24aef62fdd4cae1136cfc3f/cli/stream_command.go#L219

Ie we add the `--templates` option and default it to `true`, the system will then automatically add `--no-templates`.  You probably wouldn't need the `IsSetByUser()` part

Best in a new PR

nats/tests/nats_test.go
```diff
@@ -958,3 +958,47 @@ func TestCLIMessageRm(t *testing.T) {
 		t.Fatalf("loading delete message did not fail")
 	}
 }
+
+func TestCLIPubSendOnNewline(t *testing.T) {
```
ripienaar, line 962: mind moving this to `pub_command_test.go` like `sub_command_test.go`?

These are probably ok but would be better to set up a sub using the nats connection listening for these messages and verifying we get the right messages

thewizzy: I have made these changes. I kept the cli output only when it was part of the test, `--quiet` mode.

thewizzy:
> I think we will need to find a way to handle SIGINT before we merge this. When a user runs `natscli pub destination.subject` with the proposed PR the client doesn't terminate with ctrl+c, forcing them to kill it manually to exit.

I have implemented better signal handling that handles the IO differently. We can terminate even when the `bufio.ReadLine` is blocking, as we control the pipe.

ripienaar:
OK, looking good and ^C behaves well, please rebase and we're ok here. 

Well schedule a revisit of the whole pub/sub code cos it's a bit of a mess - I mean before this PR already so no fault of this PR.  Will also land extensive unit testing for these commands then

thewizzy:
> OK, looking good and ^C behaves well, please rebase and we're ok here.
> 
> Well schedule a revisit of the whole pub/sub code cos it's a bit of a mess - I mean before this PR already so no fault of this PR. Will also land extensive unit testing for these commands then

I noticed whilst looking at testing that the timeout in `runNatsCli` does not work - I have a potential fix for nix. The other thing I noted is that there are a number of tests that do not work on Windows, partly because the `runNatsCli` command only supports windows (or bash on windows) and a few other tests that assume a nix environment. Many can be fixed by having a `runNatsCli` specifically for windows and wraping the command in `cmd /c` - though there are still a few tests that fail. Happy to dig more into it if that is welcomed. 

It should be rebased now. Thank you.

ripienaar (approved):
LGTM

ripienaar:
Thank you @thewizzy we'll see what we can do about windows testing too /cc @ploubser

PR #1359, opened 2025-04-29
Remove redundant condition in consumer command

### LINKED ISSUE #1084: Clean up consumer code
### Proposed change

Code at line https://github.com/nats-io/natscli/blob/main/cli/consumer_command.go#L1006 is not reachable 

At line 1003  c.consumer is always != "" due to conditon at line 1002.

What is the expected behaviour?

### Use case

Fix or remove dead code

### Contribution

_No response_

Imanghvs:
Closes #1084 .

The inner conditional statement is always true because  it's part of the outer conditional statement

### DIFF
--- a/cli/consumer_command.go
+++ b/cli/consumer_command.go
@@ -1555,11 +1555,7 @@ func (c *consumerCmd) prepareConfig() (cfg *api.ConsumerConfig, err error) {
 		}
 
 		if cfg.Durable != "" && c.consumer != "" && cfg.Durable != c.consumer {
-			if c.consumer != "" {
-				cfg.Durable = c.consumer
-			} else {
-				return cfg, fmt.Errorf("durable consumer name in %s does not match CLI consumer name %s", c.inputFile, c.consumer)
-			}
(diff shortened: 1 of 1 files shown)


cli/consumer_command.go
```diff
@@ -1555,11 +1555,7 @@ func (c *consumerCmd) prepareConfig() (cfg *api.ConsumerConfig, err error) {
 		}
 
 		if cfg.Durable != "" && c.consumer != "" && cfg.Durable != c.consumer {
-			if c.consumer != "" {
-				cfg.Durable = c.consumer
-			} else {
-				return cfg, fmt.Errorf("durable consumer name in %s does not match CLI consumer name %s", c.inputFile, c.consumer)
-			}
+			cfg.Durable = c.consumer
```
ripienaar, line 1558: we would still need to return the error in a else though right for when both are set but different

ripienaar: actually never mind, I am being dumb.

ripienaar (approved):
LGTM

PR #1360, opened 2025-04-30
Cross platform testing & fix runNatsCli timeout

thewizzy:
The tests relied on `bash -c` to take in the commands from strings, as created by `fmt.Sprintf` statements throughout the code. This isn't available on Windows, the equivalent `cmd /c` does not support `""` or `''` quotes for arguments which would break a few of the tests. 

`exec.Command()` expects to take a slice and when provided with a slice of arguments the spaces do not matter - nor is there a need for `bash -c` or `cmd /c`. 

This PR provides the following:
* Converts all command arguments from strings to slices
* Fixes the timeout, it was not working as bash was spawning go and go the program. 
* Provides an implementation for *nix variants and Windows.

I have tested it on Windows, Linux and Mac. 

Note: For Windows you need `wc` from [coreutils](https://gnuwin32.sourceforge.net/packages/coreutils.htm) in your PATH. The other platforms should have it by default.

Question: In the Command there is the following `$(ls ../*.go | grep -v _test.go)`. Is there a historical purpose of this? The test files have been removed from the folder and there is only one go file now.

### DIFF
--- a/cli/auth_nkey_command.go
+++ b/cli/auth_nkey_command.go
@@ -287,6 +287,7 @@ func (c *authNKCommand) sealAction(_ *fisk.ParseContext) error {
 	if err != nil {
 		return err
 	}
+	defer f.Close()
 
 	_, err = f.Write(encryptedData)
 	if err != nil {
@@ -342,6 +343,7 @@ func (c *authNKCommand) unsealAction(_ *fisk.ParseContext) error {
 	if err != nil {
 		return err
 	}
+	defer f.Close()
 
 	_, err = f.Write(decryptedData)
 	if err != nil {
--- a/cli/auth_xkey_test.go
+++ b/cli/auth_xkey_test.go
@@ -15,12 +15,11 @@ package cli
 
 import (
 	"crypto/rand"
+	"github.com/nats-io/nkeys"
 	"io"
 	"os"
 	"path/filepath"
 	"testing"
-
-	"github.com/nats-io/nkeys"
 )
 
 func TestSealUnseal(t *testing.T) {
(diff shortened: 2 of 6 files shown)


ploubser:
👍 

1 comment, would you consider using something like https://github.com/kballard/go-shellquote? That way we can keep the `runNatsCli` call as is and you do the splitting in the function itself?

Also to answer your question, yes `$(ls ../*.go | grep -v _test.go)` is from before we moved the tests to their own directory.

ploubser (changes requested):
👍

1 comment, would you consider using something like https://github.com/kballard/go-shellquote? That way we can keep the runNatsCli call as is and you do the splitting in the function itself?

Also to answer your question, yes $(ls ../*.go | grep -v _test.go) is from before we moved the tests to their own directory.

thewizzy:
> 👍
> 
> 1 comment, would you consider using something like https://github.com/kballard/go-shellquote? That way we can keep the `runNatsCli` call as is and you do the splitting in the function itself?
> 
> Also to answer your question, yes `$(ls ../*.go | grep -v _test.go)` is from before we moved the tests to their own directory.

Thanks Pieter. I didn't see it was already a dep, so makes sense to use it. 

The non `runNatsCli` related changes were from failing tests on windows. If there is interest we could add windows testing to the GH Action.

We can use something like the following. Would just need to include `libintl3.dll`, `libiconv2.dll` and `wc.exe` for some of the subscribe translate tests to work.
```
go list ./... |
  Where-Object { $_ -notmatch "asciigraph" } |
  ForEach-Object { go test -v --failfast -p=1 $_ }
```

ploubser:
Thanks @thewizzy. There is definitely interest in adding windows testing to the GH actions. If that is something you're interested in taking on great, otherwise I will have a look at it as soon as I have time.

PR #1362, opened 2025-05-01
Add object edit command

### LINKED ISSUE #1262: Support configuration editing of "object" and "kv" stores
### Proposed change

The edit configuration functionality is very useful. Stream and consumer configuration editing is fully supported by the command line tool. The same functionality would be good to have for the object store and the key-value store.

### Use case

Changing the replica number in an already deployed instance without losing the content of the object store or key-value store.

### Contribution

_No response_

Imanghvs:
Closes #1262 .
Edit for kv is already provided by another contribution. Inspired by that, here is the edit command for object store buckets.
Also there's a bonus commit here to sort the imports (they are already sorted in other files, so this one was probably missing).

### DIFF
--- a/cli/object_command.go
+++ b/cli/object_command.go
@@ -16,6 +16,7 @@ package cli
 import (
 	"context"
 	"encoding/base64"
+	"errors"
 	"fmt"
 	"io"
 	"os"
@@ -54,6 +55,15 @@ type objCommand struct {
 	replicas    uint
 	ttl         time.Duration
 	compression bool
+
+	ttlIsSetByUser           bool
+	replicasIsSetByUser      bool
+	maxBucketSizeIsSetByUser bool
+	descriptionIsSetByUser   bool
+	tagsIsSetByUser          bool
+	clusterIsSetByUser       bool
+	metadataIsSetByUser      bool
+	compressionIsSetByUser   bool
 }
 
 func configureObjectCommand(app commandHost) {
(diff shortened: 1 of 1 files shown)


cli/object_command.go
```diff
@@ -689,6 +698,53 @@ func (c *objCommand) addAction(_ *fisk.ParseContext) error {
...
+		Bucket:      c.bucket,
+		Description: c.description,
+		TTL:         c.ttl,
+		Storage:     nfo.Config.Storage,
+		Replicas:    int(c.replicas),
+		Placement:   placement,
+		MaxBytes:    c.maxBucketSize,
+		Metadata:    c.metadata,
+		Compression: c.compression,
+	})
```
ripienaar, line 740: Won't this mean that if `metadata` or any of these others aren't set one the CLI then these would go empty values? or default values?

Ideally you should load the existing bucket, make a new config copying values from existing config to this new config and then apply only those settings that was set on the CLI ontop - and finally pass that into Update.  

This way you only update what actually changed.

And now that I look at the KV code it has the same structural problem, so I think we'll need to revisit that one first so you can use that as a blueprint for how to do the obj one?

wdyt @ploubser

ploubser: Yes, KV has the same issue. Editing only 1 field will reset the others.

ploubser: I should be able to fix up KV next week. I'll ping on this PR when it's done.

Imanghvs: Okay thanks for the review. We don't have to wait for the kv fix. I like to push a fix for this object store today on this branch and if you like it we can then merge and also use it for kv later.

ripienaar: OK Thanks @Imanghvs 

If you need some guidance the way to do it is to use the `IsSetbyUser()`  feature of the flag parser, this lets you determine if something is set or not and so know if you should pass in that new value.  Grep for that you'll find many examples here

Imanghvs: Oh thank you so much I was getting a headache over this part

Imanghvs: I just fixed and tested. It's now waiting for another review.

ripienaar:
Thanks, will take a look soon but might be only weekend/monday

cli/object_command.go
```diff
@@ -57,6 +57,15 @@ type objCommand struct {
...
 }
 
+var ttlIsSetByUser bool
+var replicasIsSetByUser bool
+var maxBucketSizeIsSetByUser bool
+var descriptionIsSetByUser bool
+var tagsIsSetByUser bool
+var clusterIsSetByUser bool
+var metadataIsSetByUser bool
+var compressionIsSetByUser bool
```
ripienaar, line 67: add these variables into the `objCommand` please, trying to keep the pollution of the `cli` package a bit small so all the varsity attach to the command struct

ripienaar (changes requested):
Looks pretty good, just one minor request, thank you

cli/object_command.go
Imanghvs: Changes applied!

ripienaar:
@Imanghvs let me know when you're ready  - I see you are still pushing here

Imanghvs:
> @Imanghvs let me know when you're ready - I see you are still pushing here

No more pushes, I'm ready :)

ripienaar:
Thank you @Imanghvs

PR #1364, opened 2025-05-02
Add stream tests

ploubser:


nats/tests/consumer_command_test.go
```diff
@@ -204,22 +184,21 @@ func TestConsumerState(t *testing.T) {
...
-			"last_delivered_msg": "Last Delivered Message: Consumer sequence: 0 Stream sequence: 0",
-			"ack_floor":          "Acknowledgment Floor: Consumer sequence: 0 Stream sequence: 0",
-			"outstanding_acks":   "Outstanding Acks: 0 out of maximum 1,000",
-			"redelivered_msgs":   "Redelivered Messages: 0",
-			"unprocessed_msgs":   "Unprocessed Messages: 0",
-			"waiting_pulls":      "Waiting Pulls: 0 of maximum 512",
+	expected := map[string]any{
+		"Header": `State for Consumer ORDERS > TEST_\d+ created .*`,
+		"State": map[string]any{
+			"Acknowledgment Floor":   `Consumer sequence: 0 Stream sequence: 0`,
```
ripienaar, line 190: Why the mix of string quoting formats here? Other ones use `"` in other tests

ripienaar (changes requested):
minor nit pick

nats/tests/consumer_command_test.go
ploubser: To limit escape characters. When escaping I ideally want to use `` instead of "", but I see there are some inconsistencies here. Let me fix.

ripienaar (approved):
LGTM

PR #1366, opened 2025-05-02
Add Windows to GH Action Tests

thewizzy:
Tests were not working well on Windows until recently, now they are in a place where we can include them in the CI tests.

* Tests on Windows & Linux, Go 1.23 & 1.24 on each
* Linting runs first and blocks tests. Saving time.
* Windows tests will fail fast on error, similar behavior to Linux
* Windows runners have `wc` by default

Considered adding `golangci-lint` to replace the current linting, there are a lot of potential issues. A task for another day.


thewizzy:
> Looks we have 2 test runs?
I think the issue here is that the old tests are still an expectation of /this/ repository until this is merged. 
On my fork they do not run at all, nor should they as they come from the updated file.
https://github.com/thewizzy/natscli/actions/runs/14826209138

> 
> Besides that it looks good. If we can fix those tests and rebase we can merge.
GitHub can cleanly merge this as is, but can do a manual rebase if required.

ploubser (approved):
Makes sense. Looks good.

ploubser (approved):

PR #1388, opened 2025-05-21
Add tests for sub command flags

ploubser:


nats/tests/sub_command_test.go
```diff
@@ -17,30 +17,76 @@ import (
...
+
+	err = nc.PublishMsg(&msg)
+	if err != nil {
+		t.Fatalf("unable to publish message: %s", err)
+	}
+
+	return srv, nc, mgr
+}
+
+func setupSubCommandNoJetstream(t *testing.T) (*server.Server, *nats.Conn) {
```
ripienaar, line 59: feels like we need to make a few standard says to setup servers and use it in all tests.

Like, this one isnt really `sub` specific but its in the name

ripienaar (changes requested):
comment about setup

nats/tests/sub_command_test.go
ploubser: Yeah, I agree. I've extracted another generic setup function into nats_test.go and moved some of the logic into the test itself. Let me know what you think. I can update the other tests in another PR if this works for you.

ripienaar (approved):
LGTM

ripienaar (approved):
LGTM

PR #1391, opened 2025-05-26
Release using stable go

ripienaar:
This ensures we use latest go and not latest-1


bruth (approved):
Sorry missed this earlier this week.

Issue #1375, opened 2025-05-09
Support a direct get based sub command for work queues


ripienaar:
### Proposed change

The `nats sub` command, when access streams, can do so only via a consumer.

We should support walking the stream using batched direct gets instead which would mean one can sub to a Work Queue also without worrying that messages will be acked and removed

A new `--direct` flag should be added which opts into this, or if a WQ is accessed, ADR-31 based batched direct gets should be done to walk the stream then (like the jsm pager already supports).

Review pager code to see what's possible, I believe since 2.11 we can also do time range based batch gets which pager does not yet support - we can support it there also while doing this.

To use this feature the stream info should report at least version 1 api as the `_nats.created.server.api_level`

### Use case

sub to work queues and streams without consumers.

PR #1394, opened 2025-06-02
(#1375) Add --direct flag to sub command

ploubser:
Add direct to the sub command, leveraging `jetstreamext.GetBatch`.

If we detect that we are subscribed to a Work Queue direct will be set to true automatically.


cli/sub_command.go
```diff
@@ -309,6 +314,30 @@ func (c *subCmd) validateInputs(nc *nats.Conn) error {
...
 
+	isWorkQueue, err := iu.IsWorkQueue(c.stream, nc)
+	if err != nil {
+		return err
+	}
+	c.direct = c.direct || isWorkQueue
+
+	if c.direct {
+		if c.stream == "" {
+			return fmt.Errorf("cannot enable direct gets: stream name is required")
```
ripienaar, line 325: hmm, could make this work, there's a API for finding a stream hosting a subject. `jsm. StreamNames`

internal/util/util.go
```diff
@@ -639,3 +641,36 @@ func ParseStringAsBytes(s string) (int64, error) {
...
+func IsWorkQueue(streamName string, nc *nats.Conn) (bool, error) {
+	if streamName == "" {
+		return false, nil
+	}
+
+	mgr, err := jsm.New(nc)
+	if err != nil {
+		return false, err
+	}
+	stream, err := mgr.LoadOrNewStream(streamName)
```
ripienaar, line 655: should just load

ripienaar (changes requested):
I need to test this but some short comments meanwhile

internal/util/util.go
```diff
@@ -639,3 +641,36 @@ func ParseStringAsBytes(s string) (int64, error) {
 
 	return num, nil
 }
+
+// IsWorkQueue inspecs a stream's config to see if it's a workqueue or not
+func IsWorkQueue(streamName string, nc *nats.Conn) (bool, error) {
```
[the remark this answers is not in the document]
ploubser: I'll rewrite this to just create a single manager in the validate function and then do a single stream load from there too. The api level validation is happening in that function, so ideally we want to know if the stream's retention policy implies direct access, otherwise we have to fail late. 

We can totally do that though if that's what you prefer? Do the api level validation after we load the stream in the directSubscribe function.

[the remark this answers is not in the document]
ripienaar: Yeah something like that, as long as we only do info once that's ideal

cli/sub_command.go
```diff
@@ -309,6 +316,53 @@ func (c *subCmd) validateInputs(nc *nats.Conn) error {
...
+		c.streamObj, err = mgr.LoadStream(c.stream)
+		if err != nil {
+			return err
+		}
+
+		// Check if the stream is a Work Queue
+		c.direct = c.direct || c.streamObj.Retention() == api.RetentionPolicy(nats.WorkQueuePolicy)
+
+		if c.direct {
+			apiLevel, err := mgr.MetaApiLevel(true)
```
ripienaar, line 356: make this `false`, it should save some round trips

cli/sub_command.go
```diff
@@ -309,6 +316,53 @@ func (c *subCmd) validateInputs(nc *nats.Conn) error {
...
+			if err != nil {
+				return err
+			}
+			if apiLevel < 1 {
+				return fmt.Errorf("cannot enable direct gets: API level 1 is required")
```
ripienaar, line 361:
```
return fmt.Errorf("subscribing using direct get requires NATS Server 2.11")
```

cli/util.go
```diff
@@ -332,6 +332,32 @@ func jsOpts() []nats.JSOpt {
 	return jso
 }
 
+func newJetStreamWithOpts(nc *nats.Conn) (jetstream.JetStream, error) {
```
ripienaar, line 335: Isn't this basically `prepareJSHelper()`

cli/sub_command.go
```diff
@@ -630,6 +680,108 @@ func (c *subCmd) defaultSubscribe(nc *nats.Conn, handler nats.MsgHandler, subs *
 	return nil
 }
 
+func (c *subCmd) directSubscribe(subCtx context.Context, nc *nats.Conn, handler nats.MsgHandler, subs *[]*nats.Subscription, ignoreSubjects []string) error {
```
ripienaar, line 683: This should detect if a stream doesn't have direct on and fail a bit better

```
[rip@s1-lon]% bin/nats sub '>' --stream ORDERS_0 --last --direct
13:31:07 Subscribing to JetStream Stream (direct get) holding messages with subject > starting with the last message received
nats: error: nats: no responders available for request
```

cli/sub_command.go
```diff
@@ -309,6 +316,53 @@ func (c *subCmd) validateInputs(nc *nats.Conn) error {
 		}
 	}
 
+	if c.jetStream {
+		mgr, err := jsm.New(nc)
```
ripienaar, line 320: should use the existing helpers to get a mgr because those will set up tracing and stuff for you

cli/sub_command.go
```diff
@@ -630,6 +680,108 @@ func (c *subCmd) defaultSubscribe(nc *nats.Conn, handler nats.MsgHandler, subs *
...
+
+		if err != nil {
+			if errors.Is(err, context.Canceled) {
+				return nil
+			}
+			return err
+		}
+
+		for msg, err := range msgs {
+			if errors.Is(err, jetstreamext.ErrNoMessages) {
```
ripienaar, line 761: maybe better as a case?

PR #1402, opened 2025-06-13
Initial support for atomic batches

ripienaar:


ripienaar:
Blocked on https://github.com/nats-io/jsm.go/pull/668

Issue #1393, opened 2025-05-28
Downgrade safety report


ripienaar:
### Proposed change

Since 2.11 server we have the concept of API level and jetstream assets report if they are using 2.11 features by requiring api level 1

We need to make a tool that walk JSZ output and reports all streams and consumers that will not be safe to start.

Argument should be the proposed new max api level supported and then output is all assets that won't work in a nice table style output.

Additionally, if not already done, `stream find` and `consumer find` should be able to find on api level

### Use case

Safe downgrades

PR #1404, opened 2025-06-16
(#1393) Add downgrade safety report

ploubser:
This commit introduces a new CLI command: `report downgrade`.

Running `nats report downgrade <api-level>` checks the required API levels for streams and consumers, and reports which assets would fail to start if the server were downgraded to the specified API level.

By default, only consumers that are directly incompatible are shown. To include consumers that inherit incompatibility from their streams, use the `--all` flag.

Output is shown in a table format by default. `--json` is also included.

### DIFF
--- a/cli/server_report_command.go
+++ b/cli/server_report_command.go
@@ -23,6 +23,7 @@ import (
 	"syscall"
 	"time"
 
+	"github.com/nats-io/natscli/internal/sysclient"
 	iu "github.com/nats-io/natscli/internal/util"
 
 	"github.com/choria-io/fisk"
@@ -59,6 +60,8 @@ type SrvReportCmd struct {
 	consumer                string
 	watchInterval           int
 	nc                      *nats.Conn
+	apiLevel                uint
+	all                     bool
 }
 
 type srvReportAccountInfo struct {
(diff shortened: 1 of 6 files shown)


cli/util.go
```diff
@@ -896,3 +896,11 @@ func calculateRate(new, last float64, since time.Duration) float64 {
 
 	return (new - last) / since.Seconds()
 }
+
+func parseApiLevel(lvl string) uint {
```
ripienaar, line 900: move to internal/util - sorry should have been clearer since there are multiple utils, trying to minimise noise in the cli package

cli/server_report_command.go
```diff
@@ -1540,3 +1548,118 @@ func (c *SrvReportCmd) reqFilter() server.EventFilterOptions {
...
+	return nil
+}
+
+func (c *SrvReportCmd) renderDowngrade(as jsDowngradeAsset) {
+	if !c.json {
+		c.printDowngradeReport(as)
+		return
+	}
+
+	if output, err := iu.ToJSON(as); err != nil {
```
ripienaar, line 1633: iu.PrintJSON()

cli/server_report_command.go
```diff
@@ -1540,3 +1548,118 @@ func (c *SrvReportCmd) reqFilter() server.EventFilterOptions {
...
+		return fmt.Errorf("connection setup failed: %v", err)
+	}
+	defer nc.Close()
+
+	if !c.json {
+		fmt.Println("Obtaining JetStream Asset information")
+	}
+
+	sys := sysclient.New(nc)
+	accounts, err := sys.CollectClusterAccounts(opts().Timeout)
```
ripienaar, line 1582: timeout doesn't seem to be handled right

```
[rip@s1-lon]% bin/nats srv report downgrade 0 --context system.lon --all --timeout 1ms
Obtaining JetStream Asset information
All assets are compatible with the specified API level.
```

when in fact there are - fetching just timed out, some timeout error isnt propagating up

internal/sysclient/sysclient.go
```diff
@@ -87,48 +89,162 @@ func (s *SysClient) JszPing(opts server.JszEventOptions, fopts ...FetchOpt) ([]J
...
+
 func (s *SysClient) Fetch(subject string, data []byte, opts ...FetchOpt) ([]*nats.Msg, error) {
 	if subject == "" {
 		return nil, fmt.Errorf("%w: expected subject 0", ErrValidation)
 	}
 
 	conn := s.nc
-	reqOpts := &FetchOpts{}
+	reqOpts := &FetchOpts{
+		Timeout:     5 * time.Second,
```
ripienaar, line 201: seems to ignore the supplied timeout, each request should use the supplied I think

ripienaar (changes requested):
looking good so far, some minor issues

Screenshot 2025-07-09 at 13 15 44

ripienaar (approved):
LGTM

PR #1405, opened 2025-06-25
Add --filter flag to bench js consume and fetch commands

mikluko:
## Summary
- Adds `--filter` flag to `nats bench js consume` and `nats bench js fetch` commands
- Allows filtering JetStream messages by subjects during benchmarking
- Maintains consistency with existing `--filter` flag in consumer commands

## Changes
- Added `filterSubjects []string` field to `benchCmd` struct
- Added flag definition using `StringsVar` pattern in `addJSConsumerFlags`
- Updated `createOrUpdateConsumer` to include `FilterSubjects` in consumer config
- Added filter display to benchmark banner output

### DIFF
--- a/cli/bench_command.go
+++ b/cli/bench_command.go
@@ -68,6 +68,7 @@ type benchCmd struct {
 	randomizeGets        int
 	payloadFilename      string
 	hdrs                 []string
+	filterSubjects       []string
 }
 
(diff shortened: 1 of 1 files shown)


ripienaar:
Thank you @mikluko

PR #1415, opened 2025-07-17
Only show negation options for edit operations

ripienaar:
This shows --allow-batch in stream add but in edit shows --[no-]allow-batch in this case the negation option is shown even when default is false because they exist to disable configuration features if set

### DIFF
--- a/cli/stream_command.go
+++ b/cli/stream_command.go
@@ -201,12 +201,21 @@ func configureStreamCommand(app commandHost) {
 		f.Flag("tag", "Place the stream on servers that has specific tags (pass multiple times)").IsSetByUser(&c.placementTagsSet).StringsVar(&c.placementTags)
 		f.Flag("tags", "Backward compatibility only, use --tag").Hidden().IsSetByUser(&c.placementTagsSet).StringsVar(&c.placementTags)
 		f.Flag("cluster", "Place the stream on a specific cluster").IsSetByUser(&c.placementClusterSet).StringVar(&c.placementCluster)
-		f.Flag("ack", "Acknowledge publishes").Default("true").BoolVar(&c.ack)
+		if edit {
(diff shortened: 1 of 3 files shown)


cli/stream_command.go
```diff
@@ -219,13 +228,23 @@ func configureStreamCommand(app commandHost) {
...
+			f.Flag("deny-purge", "Deny entire stream or subject purges via the API").IsSetByUser(&c.denyPurgeSet).BoolVar(&c.denyPurge)
+			f.Flag("allow-direct", "Allows fast, direct, access to stream data via the direct get API").IsSetByUser(&c.allowDirectSet).BoolVar(&c.allowDirect)
+			f.Flag("allow-mirror-direct", "Allows fast, direct, access to stream data via the direct get API on mirrors").IsSetByUser(&c.allowMirrorDirectSet).BoolVar(&c.allowMirrorDirect)
+		} else {
+			f.Flag("allow-batch", "Allow atomic batch publishing").IsSetByUser(&c.allowAtomicBatchIsSet).UnNegatableBoolVar(&c.allowAtomicBatch)
+			f.Flag("allow-counter", "Configures the Stream as a distributed counter").IsSetByUser(&c.allowCounterIsSet).UnNegatableBoolVar(&c.allowCounter)
+			f.Flag("allow-rollup", "Allows roll-ups to be done by publishing messages with special headers").IsSetByUser(&c.allowRollupSet).UnNegatableBoolVar(&c.allowRollup)
+			f.Flag("deny-delete", "Deny messages from being deleted via the API").IsSetByUser(&c.denyDeleteSet).UnNegatableBoolVar(&c.denyDelete)
+			f.Flag("deny-purge", "Deny entire stream or subject purges via the API").IsSetByUser(&c.denyPurgeSet).BoolVar(&c.denyPurge)
```
ripienaar, line 244: cant do this one as it breaks existing scripts

PR #1423, opened 2025-07-23
Use default timeout for API calls

ripienaar:
This moves to the new default timeout pattern in the jetstream package, the outcome would be that API calls time out on the default but things like watch that are long running functions will continue indefinitely,

This more or less recreate the behaviour of the legacy package


alexbozhenko:
@ripienaar thank you for the fix!
Should the same fix be applied to obj command as well?
Looks like it was `context.WithTimeout(ctx, opts().Timeout)` was introduced there at the same time with kv command:
https://github.com/nats-io/natscli/pull/1164/files#diff-bec8a7425b2c47ae9a6221136ac54711d92942f141f650821fb52d9397428971R474-R477

ripienaar:
Would have to hear what @piotrpio thinks wrt the state of the object API

piotrpio:
Yeah, object API will behave the same way as kv API in terms of timing out when listing keys etc. So I think it makes sense to have it mirrored in both cli commands, I will apply the necessary fixes in go client for both APIs as well.

ripienaar:
Thanks @piotrpio, have done object also then

piotrpio (approved):
LGTM!

PR #1428, opened 2025-07-25
Ensure parsed ints are in correct limits

ripienaar:
We have a few places where we do int32(parsedInt) but was always parsing all ints as 64 bit so that was not safe to do.

This makes sure all parsed ints are done at the right bit sizes


ploubser (approved):
LGTM

PR #1437, opened 2025-07-31
Rewrite sub to use nats.go/jetstream

ploubser:
Here we update the sub action to use the modern jetstream package. As a side effect of this we now support both pull and push consumers with the --durable flag.

### DIFF
--- a/cli/sub_command.go
+++ b/cli/sub_command.go
@@ -28,9 +28,7 @@ import (
 	"time"
 
 	"github.com/choria-io/fisk"
-	"github.com/dustin/go-humanize"
 	"github.com/nats-io/jsm.go"
-	"github.com/nats-io/jsm.go/api"
 	"github.com/nats-io/nats-server/v2/server"
 	"github.com/nats-io/nats.go"
 	"github.com/nats-io/nats.go/jetstream"
@@ -73,7 +71,7 @@ type subCmd struct {
 	height                int
 	messageRates          map[string]*subMessageRate
 	direct                bool
-	streamObj             *jsm.Stream
+	streamObj             jetstream.Stream
 }
 
 type subMessageRate struct {
@@ -81,21 +79,52 @@ type subMessageRate struct {
 	rates     []float64
 }
 
+// subscriptionState holds the shared, mutable state used during command execution
+type subscriptionState struct {
+	// msgMu guards concurrent access to message counters and printing logic
+	msgMu *sync.Mutex
+
+	// cancelFn is called to terminate sub
+	cancelFn context.CancelFunc
+
+	// counter tracks how many messages have been received across all subjects
+	counter uint
+
+	// startTime is the timestamp at which the command execution started
+	startTime time.Time
+
+	// subjMu guards concurrent access to the subject reporting maps
+	subjMu *sync.Mutex
+
+	// ignoreSubjects is the list of subjects to skip when receiving messages
+	ignoreSubjects []string
+
+	// subjectReportMap tracks how many messages were received per subject
+	subjectReportMap map[string]int64
+
+	// subjectBytesReportMap tracks how many bytes were received per subject
+	subjectBytesReportMap map[string]int64
+
+	// matchMap holds inbound request messages, keyed by their expected reply subject
+	matchMap map[string]*nats.Msg
+
+	// dump determines whether messages should be dumped to stdout or files
+	dump bool
+}
+
 func configureSubCommand(app commandHost) {
 	c := &subCmd{}
 
 	subHelp := `
-	Jetstream will be activated when related options like --stream, --durable or --ack are supplied.
+	Jetstream will be activated when related options like --stream, --durable, direct or --ack are supplied.
 
 		E.g. nats sub <subject that is bound to a stream> --all
 
-	Currently only supports push subscriptions. Uses an ephemeral consumer without ack by default.  
+	Uses an ephemeral consumer without ack by default.  
 
 	For specific consumer options please pre-create a consumer using 'nats consumer add'.app.
 
 		E.g. when explicit acknowledgement is required.
-
-	Caution: Be careful when subscribing to streams with WorkQueue policy. Messages will be acked and deleted when a durable consumer is being used.
 
 	Use nats stream view <stream> for inspecting messages.	
 		
@@ -111,12 +140,8 @@ func configureSubCommand(app commandHost) {
 	act.Flag("raw", "Show the raw data received").Short('r').UnNegatableBoolVar(&c.raw)
 	act.Flag("translate", "Translate the message data by running it through the given command before output").StringVar(&c.translate)
 	act.Flag("ack", "Acknowledge JetStream message that have the correct metadata").BoolVar(&c.jsAck)
-	// We do not support (explicit) ackPolicy right now. The only situation where it is useful would be WorkQueue policy right now.
-	// Deleting from a stream with WorkQueue through ack could be unexpected behavior in the sub command.
-	// To be done - check for streams with WorkQueue, then prompt with or allow with override --force=WorkQueueDelete
-	// act.Flag("ackPolicy", "Acknowledgment policy (none, all, explicit) (requires JetStream)").Default("none").EnumVar(&c.ackPolicy, "none", "all", "explicit")
 	act.Flag("match-replies", "Match replies to requests").UnNegatableBoolVar(&c.match)
-	act.Flag("inbox", "Subscribes to a generate inbox").Short('i').UnNegatableBoolVar(&c.inbox)
+	act.Flag("inbox", "Subscribes to a generated inbox").Short('i').UnNegatableBoolVar(&c.inbox)
 	act.Flag("count", "Quit after receiving this many messages").UintVar(&c.limit)
 	act.Flag("dump", "Dump received messages to files, 1 file per message. Specify - for null terminated STDOUT for use with xargs -0").PlaceHolder("DIRECTORY").StringVar(&c.dump)
 	act.Flag("headers-only", "Do not render any data, shows only headers").UnNegatableBoolVar(&c.headersOnly)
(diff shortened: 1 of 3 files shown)


cli/sub_command.go
```diff
@@ -317,47 +316,55 @@ func (c *subCmd) validateInputs(nc *nats.Conn, mgr *jsm.Manager) error {
...
-			if !c.streamObj.DirectAllowed() {
+			if !c.streamObj.CachedInfo().Config.AllowDirect {
 				return fmt.Errorf("cannot enable direct gets: allow_direct is not set to true for stream %s", c.stream)
 			}
 			err := iu.RequireAPILevel(mgr, 1, "subscribing using direct get requires NATS Server 2.11")
 			if err != nil {
 				return err
 			}
 		}
+
```
ploubser, line 356: I'm not sure about these. We could probably add at least some of this functionality.

cli/sub_command.go
```diff
@@ -560,100 +626,152 @@ func (c *subCmd) reportSubscribe(ctx context.Context, nc *nats.Conn, handler nat
...
+	if err == jetstream.ErrNotPullConsumer {
+		pcons, err := js.PushConsumer(ctx, c.stream, c.durable)
+		if err != nil {
+			return nil, nil, nil, true, err
+		}
+		info, err := pcons.Info(ctx)
+		if err != nil {
+			return nil, nil, nil, true, err
+		}
+		log.Printf("Subscribing to JetStream Stream %q using existing push consumer %q", c.stream, c.durable)
```
ploubser, line 750: Do we want to log a deprecation notice here?

cli/sub_command.go
ripienaar: about needing 2.11? I am ok with that, good things come to those who upgrade

ripienaar:
We should figure out how to get trace working here:

```
[rip@s1-lon]% bin/nats sub '>' --stream WQ --trace                                                                                      <14:01:52
14:01:54 >>> Connected to nats://n2-lon:4222 (138.68.149.96:4222)
14:01:54 >>> $JS.API.INFO


14:01:54 <<< $JS.API.INFO
{"type":"io.nats.jetstream.api.v1.account_info_response","memory":0,"storage":6188433794,"reserved_memory":0,"reserved_storage":1073741824,"stream
s":125,"consumers":511,"limits":{"max_memory":-1,"max_storage":-1,"max_streams":-1,"max_consumers":-1,"max_ack_pending":-1,"memory_max_stream_byte
s":-1,"storage_max_stream_bytes":-1,"max_bytes_required":false},"domain":"hub","api":{"level":2,"total":4337,"errors":7}}

[#1] Received JetStream message: consumer: WQ >  / subject: js.in.wq.1 / delivered: 0 / consumer seq: 0 / stream seq: 1
Nats-Subject: js.in.wq.1
Nats-Sequence: 1
```

People should see the requests and replies of direct get, I know nats.go has tracing support but might be its not complete?

ripienaar:
Looks like `--last-per-subject` with `--direct` isnt doing the right thing but cant trace (see above) so hard to say what's up.

ripienaar:
When doing direct we need to print jetstream status a bit different:

```
[#7] Received JetStream message: consumer: ORDERS_1 >  / subject: js.in.orders_1 / delivered: 0 / consumer seq: 0 / stream seq: 44947620
```

There's no consumer or delivered related stuff - its fetching direct from the stream - so we should not show those

cli/sub_command.go
```diff
@@ -317,47 +316,44 @@ func (c *subCmd) validateInputs(nc *nats.Conn, mgr *jsm.Manager) error {
...
-		var err error
-		c.streamObj, err = mgr.LoadStream(c.stream)
+		c.streamObj, err = js.Stream(ctx, c.stream)
 		if err != nil {
 			return err
 		}
 
 		// Check if the stream is a Work Queue
-		c.direct = c.direct || c.streamObj.Retention() == api.RetentionPolicy(nats.WorkQueuePolicy)
+		c.direct = c.direct || c.streamObj.CachedInfo().Config.Retention == jetstream.WorkQueuePolicy
```
ripienaar, line 337: should probably also do for interest retention

cli/sub_command.go
```diff
@@ -317,47 +316,44 @@ func (c *subCmd) validateInputs(nc *nats.Conn, mgr *jsm.Manager) error {
...
 
 		if c.direct {
+			if len(c.subjects) > 1 {
+				return fmt.Errorf("cannot enable direct gets: cannot perform direct gets from multiple subjects")
```
ripienaar, line 341: we kind of can now, so maybe open a follow up issue to circle back to that

cli/sub_command.go
```diff
@@ -422,9 +418,58 @@ func (c *subCmd) createMsgHandler(hcfg msgHandlerState, nc *nats.Conn, subjMu *s
 	}
 }
 
-func (c *subCmd) shouldIgnore(m *nats.Msg, ignoreSubjects []string) bool {
+func (c *subCmd) createJestreamMsgHandler(hcfg msgHandlerState, nc *nats.Conn, subjMu *sync.Mutex, consumers *[]jetstream.ConsumeContext, ignoreSubjects []string, subjectReportMap map[string]int64, subjectBytesReportMap map[string]int64, matchMap map[string]*nats.Msg, t *time.Timer, dump bool) jetstream.MessageHandler {
```
ripienaar, line 421: var here `hcfg` seems a bit like handler config, can we perhaps pass some of these things in via that instead?  

Otherwise at least should comment what this stuff all does, we'll be hating ourselves in the future

otoh the other also have these giant arg lists so probably just comment it

cli/sub_command.go
```diff
@@ -422,9 +418,58 @@ func (c *subCmd) createMsgHandler(hcfg msgHandlerState, nc *nats.Conn, subjMu *s
...
+		if c.jsAck {
+			defer func() {
+				err := m.Ack()
+				if err != nil && !dump && !c.raw {
+					log.Printf("Acknowledging message via subject %s failed: %s\n", m.Reply, err)
+				}
+			}()
+		}
+
+		*hcfg.counter++
```
ripienaar, line 443: wonder why this is a pointer, we dont seem to really use its pointer abilities?

cli/sub_command.go
```diff
@@ -445,6 +490,20 @@ func (c *subCmd) handleFlowControlMsg(m *nats.Msg, nc *nats.Conn) bool {
 	return false
 }
 
+func (c *subCmd) handleJetstreamFlowContolMsg(m jetstream.Msg, nc *nats.Conn) bool {
+	if c.jetStream && len(m.Data()) == 0 && m.Headers().Get("Status") == "100" {
```
ripienaar, line 494: would you ever get a jetstream.Msg if c.jetStream isnt true?

cli/sub_command.go
```diff
@@ -445,6 +490,20 @@ func (c *subCmd) handleFlowControlMsg(m *nats.Msg, nc *nats.Conn) bool {
...
+		if m.Reply() != "" {
+			_ = nc.Publish(m.Reply(), nil)
+			log.Printf("Responding to Flow Control message")
+		} else if stalled := m.Headers().Get("Nats-Consumer-Stalled"); stalled != "" {
+			_ = nc.Publish(stalled, nil)
```
ripienaar, line 499: not a huge fan of the `_` assignment

cli/sub_command.go
```diff
@@ -560,100 +627,160 @@ func (c *subCmd) reportSubscribe(ctx context.Context, nc *nats.Conn, handler nat
 	return nil
 }
 
-func (c *subCmd) jetStreamSubscribe(nc *nats.Conn, handler nats.MsgHandler, subs *[]*nats.Subscription, ignoreSubjects []string) error {
-	js, err := nc.JetStream(jsOpts()...)
-	if err != nil {
-		return err
+func (c *subCmd) jetStreamSubscribe(ctx context.Context, handler jetstream.MessageHandler, consumerContexts *[]jetstream.ConsumeContext, ignoreSubjects []string, js jetstream.JetStream) error {
```
ripienaar, line 630: also comment these please

cli/sub_command.go
```diff
@@ -560,100 +627,160 @@ func (c *subCmd) reportSubscribe(ctx context.Context, nc *nats.Conn, handler nat
...
 	}
+	*consumerContexts = append(*consumerContexts, cCtx)
 
-	ignoredSubjInfo := ""
-	if len(ignoreSubjects) > 0 {
-		ignoredSubjInfo = fmt.Sprintf("\nIgnored subjects: %s", f(ignoreSubjects))
+	return nil
+}
+
+func (c *subCmd) reportJetstreamSubscribe(ctx context.Context, handler jetstream.MessageHandler, consumerContexts *[]jetstream.ConsumeContext, ignoreSubjects []string, js jetstream.JetStream, subs *[]*nats.Subscription, subjMu *sync.Mutex, subjectReportMap map[string]int64, subjectBytesReportMap map[string]int64) error {
```
ripienaar, line 683: commment these

cli/util.go
```diff
@@ -311,6 +311,39 @@ func jetstreamOpts() []jetstream.JetStreamOpt {
 	return res
 }
 
+// New function doesn't set a timeout. New jetstream API uses contexts for timeouts
```
ripienaar, line 314: hmm, there is a default timeout that would be used all over for timeouts when no context is given, might be worth setting that just from opts.Timeout to be safe

ripienaar (changes requested):
getting there, but you're right does feel a bit like moving deck chairs around but I think getting on nats.go here is valuable

ripienaar (changes requested):
Few small things:

 * Remove comments and disabled flag around https://github.com/nats-io/natscli/blob/d79091d2698c9b2b0d23384284b9d12869f30ed4/cli/sub_command.go#L146
 * Replace calls to humanize.IBytes with fiBytes()
 * Double check the subHelp text still make sense and fix if needed

cli/sub_command.go
```diff
@@ -317,127 +349,170 @@ func (c *subCmd) validateInputs(nc *nats.Conn, mgr *jsm.Manager) error {
...
-		c.streamObj, err = mgr.LoadStream(c.stream)
+		c.streamObj, err = js.Stream(ctx, c.stream)
 		if err != nil {
 			return err
 		}
 
+		config := c.streamObj.CachedInfo().Config
 		// Check if the stream is a Work Queue
-		c.direct = c.direct || c.streamObj.Retention() == api.RetentionPolicy(nats.WorkQueuePolicy)
+		c.direct = c.direct || config.Retention == jetstream.WorkQueuePolicy || config.Retention == jetstream.InterestPolicy
```
ripienaar, line 371: fix comment here - or just remove it, its clear as is

Issue #1387, opened 2025-05-20
Ability to pass a priority group for nats sub


jnmoyne:
### Proposed change

Currently you can't consume messages in any way if the consumer has a priority group defined.

### Use case

Being able to get messages from a consumer that has a priority group set

ripienaar:
Good idea. We are currently rewriting sub so it’s easier to maintain and has tests etc. 


Sometning to be done after that.

ddemirxxo:
As far as I know my partner only knows interment details

PR #1449, opened 2025-08-07
(#1387) Add priority groups flags to sub

ploubser:
In the past we were unable to receive and messages via durable consumers if priority groups were set.

This flag allows us to specify a priority-group with a durable consumer.

### DIFF
--- a/nats/tests/sub_command_test.go
+++ b/nats/tests/sub_command_test.go
@@ -776,6 +776,37 @@ func TestJetStreamSubscribe(t *testing.T) {
 		})
 	})
 
+	t.Run("--durable with priority groups", func(t *testing.T) {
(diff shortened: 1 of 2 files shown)


nats/tests/sub_command_test.go
```diff
@@ -776,6 +776,37 @@ func TestJetStreamSubscribe(t *testing.T) {
...
+
+			err = nc.PublishMsg(defaultTestMsg)
+			if err != nil {
+				t.Fatalf("unable to publish message: %s", err)
+			}
+
+			ctx := context.Background()
+			_, err = js.CreateConsumer(ctx, "TEST_STREAM", jetstream.ConsumerConfig{
+				Durable:        "TEST_PULL",
+				PriorityGroups: []string{"TEST_GROUP"},
```
ripienaar, line 795: this is essentially going to be ignored without setting a policy

PR #1457, opened 2025-08-13
Validate history falls within limits

ripienaar:
Previously picking --history 1000 would cast without validation to int8() and result in history 16, now we validate its within the specifically allowed ranges


cli/kv_command.go
```diff
@@ -99,7 +99,7 @@ for an indefinite period or a per-bucket configured TTL.
 	addCreateFlags := func(f *fisk.CmdClause, edit bool) {
 		f.Arg("bucket", "The bucket to act on").Required().StringVar(&c.bucket)
 		f.Flag("description", "A description for the bucket").IsSetByUser(&c.descriptionSet).StringVar(&c.description)
-		f.Flag("history", "How many historic values to keep per key").IsSetByUser(&c.historySet).Default("1").Uint64Var(&c.history)
+		f.Flag("history", "How many historic values to keep per key").IsSetByUser(&c.historySet).Default("1").Validator(iu.Int64RangeValidator(0, 64)).Int64Var(&c.history)
```
ploubser, line 102: Should this be `iu.Int64RangeValidator(1, 64)`?

ripienaar: yes, doh - though the client library would also catch it, but good catch

ploubser (approved):
👍

PR #1465, opened 2025-08-20
Remove API Errors from jsz reports

ripienaar:
This stat is misleading and does not help anyone diagnose the problems reported.


ploubser (approved):
LGTM

PR #1469, opened 2025-08-21
Update server_check_command.go

roeschter:
The usage of thresholds was not quite clearly documented
Possibility to have multiple thresholds was not described
Inversion of warn and critical was not documented, works for almost all values. 
Some units were wrong.
Some help entries were wrong


ploubser (approved):
👍

PR #1474, opened 2025-08-29
Batched direct gets in `nats bench`

jnmoyne:
Adds a new `js get` bench command that uses batched direct gets to retrieve the messages from the stream
Adds the client number to the log printouts in bench
Adds --filter to bench js ordered


ripienaar (approved):
LGTM, lets also hear from @MauriceVanVeen

MauriceVanVeen (approved):
LGTM

Issue #1478, opened 2025-09-04
nats server info returns info for wrong server


camphillips22:
### Observed behavior

When running `nats server info <name>`, the event filter options uses `string.Contains()` to filter the request. This means that if you have servers in a stateful set, and have more than 10 nodes, `nats server info nats-1` can return any of the infos for `nats-1`, `nats-10`, `nats-11`, etc.

### Expected behavior

Running `nats server info nats-1` should return the info for `nats-1`, not `nats-1X`. 

### Server and client version

nats-server: 2.11.6
nats cli: 0.2.0

### Steps to reproduce

Have one server with a hostname that contains another node's hostname. Run `nats server info nats-1` and the results will not be deterministic.

ripienaar:
Hmm, difficult one to fix, its a breaking change on the server and on this side we'd maybe sometimes have to wait for more than 1 response but we have no way to know if thats the case.

https://github.com/nats-io/nats-server/blob/82054beb61f2ee5fc8ac7a9826ec7df3fb6edc2e/server/events.go#L2052-L2062

camphillips22:
My workaround now is to use `nats server ls -j` and parse the json to get something deterministic. This works fine for what I need, but this was definitely confusing when I was getting different results each time I ran this command.

PR #1480, opened 2025-09-08
Perform exact matching when accessing z endpoints

ripienaar:
Fixes https://github.com/nats-io/natscli/issues/1478 for NATS Server 2.12 and newer


ploubser (approved):
👍

PR #1488, opened 2025-09-16
Redo bench stats and bench js get

jnmoyne:
Bench command now has its own modified and improved internal functionality for displaying and saving stats
Additional average latency stats added.
Rework bench js get.

Signed-off-by: Jean-Noël Moyne <jnmoyne@gmail.com>

### DIFF
--- a/cli/bench_command.go
+++ b/cli/bench_command.go
@@ -16,6 +16,8 @@ package cli
 import (
 	"bytes"
 	"context"
+	"encoding/csv"
+	"errors"
 	"fmt"
 	"iter"
 	"math"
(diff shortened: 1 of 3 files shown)


cli/bench_command.go
```diff
@@ -80,6 +84,36 @@ type benchCmd struct {
 	filterSubject        string   // used by JS get command
 }
 
+// A sample for a particular client
+type sample struct {
```
ripienaar, line 88: Please move these to an internal package. 

With all the many different commands in CLI package it’s hard to, for example, know what a “sample” is. 

But if they were in internal/bench/sample.go it would be much clearer what’s what and CLI tab completion in here would be cleaner for other commands

Issue #1439, opened 2025-07-31
Make Y-axis labels scale on graphs more obvious?


alexbozhenko:
### Proposed change

When looking at graphs with customers, people often find the Y axis labels confusing:
<img width="224" height="588" alt="Image" src="https://github.com/user-attachments/assets/69161ab9-8ccb-4f07-9cd4-faae110329c8" />

E.g. Is the first number 36 thousand here?

That's becuase 
https://en.wikipedia.org/wiki/Decimal_separator
is locale/culture specific.
https://randombits.dev/articles/number-localization/locale-list

Should we choose space for grouping symbol?

These two ambiguous:
```
120,000,000,000.00
120.000.000.000,00
```
This one is less so:
``` 
120 000 000 000,00
```

Alternatively, we can choose to always display the decimal part, e.g. `0.000` ?

### Use case

- "Wait, is it 15 thousand consumer create requests per second???" -
- "Oh, no. It is 15"

### Contribution

I can.

PR #1491, opened 2025-09-17
force number of decimal places on Y axis on graphs

alexbozhenko:
The function that we use to format floats on graph axis 
https://github.com/dustin/go-humanize/blob/v1.0.1/comma.go#L84
drops decimal part. That creates Y-axis labels that are hard to read.
It also causes the position of the axis to readjust as you start at the graph.

In once console:
```
nats bench pub yo --msgs=100000000 --sleep 0.01s
```
In second console:
```
nats  sub 'yo' --graph
```

image

With this change, we always display decimal part, and it makes it easier to parse:
image

Also, drop number of decimal places to two, since that high precision is barely useful in cli graphs.

Fixes: https://github.com/nats-io/natscli/issues/1439

### DIFF
--- a/cli/columns.go
+++ b/cli/columns.go
@@ -39,6 +39,10 @@ func f(v any) string {
 	return columns.F(v)
 }
 
+func fFloatFixedDecimal(v any) string {
+	return humanize.FormatFloat("#,###.##", v.(float64))
+}
+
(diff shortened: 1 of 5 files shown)


ripienaar:
Thanks

Issue #1502, opened 2025-09-23
Support batch publish


ripienaar:
We should support `--atomic` on `nats pub` that should be supported when reading from STDIN and send on new line is set.

The behaviour will be that it opens a batch and just keeps adding to the batch till done and commits the last message.

Use the batch publisher from https://github.com/synadia-io/orbit.go/tree/main/jetstreamext#batchpublisher

PR #1510, opened 2025-09-25
(#1502) Add --atomic flag to pub command

ploubser:


cli/pub_command.go
```diff
@@ -264,6 +270,75 @@ func (c *pubCmd) doReq(nc *nats.Conn, progress *progress.Tracker) error {
...
+	if !stream.CachedInfo().Config.AllowAtomicPublish {
+		return fmt.Errorf("atomic publishing is not allowed for stream %s with subject %s", streamName, c.subject)
+	}
+
+	batch, err := jetstreamext.NewBatchPublisher(js)
+	if err != nil {
+		return err
+	}
+
+	for i := 0; i < len(c.atomicPending)-1; i++ {
```
ripienaar, line 297: using range here might be less awkward

cli/pub_command.go
```diff
@@ -379,6 +454,18 @@ func (c *pubCmd) publish(_ *fisk.ParseContext) error {
 	if c.cnt < 1 {
 		c.cnt = math.MaxInt16
 	}
+	if c.atomic {
+		if !c.jetstream || !useStdin || c.sendOn != sendOnNewline {
```
ripienaar, line 458: `if !(c.jetstream && useStdin && c.sendon == sendOnNewline) {` reads a lot better

cli/pub_command.go
```diff
@@ -403,6 +490,13 @@ func (c *pubCmd) publish(_ *fisk.ParseContext) error {
 						eof = true
 					}
 					if body == "" && eof {
+						if c.atomic {
```
ripienaar, line 493: seems a bit nasty to put it here, we could add a new `if c.atomic` above line 525 and do all the atomic stuff there rather than squeeze it into the jetstream check and worse into `doJetstream`, doing it there in its own func means we can just put all the atomic things together and branch in that function around `eof` being true?

ripienaar (changes requested):
while you're there also move the tracker setup just above the `for`

cli/pub_command.go
ploubser: Moving the if atomic check out of the jetstream check is a good idea. Should have cleaned this up when I changed approach.

Annoyingly the `commit` has to happen on EOF or in the select. I'm going to move it to the select because I think it's going to be less nested noise, and let's see how it looks.

PR #1519, opened 2025-10-02
Add `--replies-all` to `nats request` for unknown number of replies

joeriddles:
Currently with `--replies` you can configure the number of expected replies to a request. If you do not know the number of expected replies, you can set this number to something high, like 100. Two downsides to this approach are that 1) that number may still be less than the number of replies and 2) if the number of replies is lower, like 99, the CLI will need to wait for the reply timeout.

`--replies-all` solves both of the above problems by allowing the responder to indicate the number of replies it will send via a header, `NATS-Replies-Total`.

### DIFF
--- a/cli/pub_command.go
+++ b/cli/pub_command.go
@@ -37,8 +37,9 @@ import (
 )
 
 const (
-	sendOnEOF     = "eof"
-	sendOnNewline = "newline"
+	sendOnEOF       = "eof"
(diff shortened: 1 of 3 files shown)


ripienaar:
Ping Tomasz, we can certainly add a strategy to Orbit its not a bad idea.

CLI can use the request-many to surface the standard ones we have

Issue #1527, opened 2025-10-14
nats server report jetstream reports incomplete or wrong RAFT meta group in leaf nodes connected with SYS account


roeschter:
### Observed behavior

When two cluster with diffrent domains are leaf node connected via a system account  $SYS.REQ.SERVER.PING.JSZ  will correctly return information from all nodes. AS cann be verified easily with --trace

The Jetstream summary will correctly list ALL nodes across two cluster but **will only list ONE RAFT Meta Group Information.**  Also when limiting th output with the filter --cluster, NO RAFT MEta Group will be shown at all.

This is somewhat annoying and confusing and makes it very hard to debug connected cluster with this command (short of looking at the raw data of $SYS.REQ.SERVER.PING.JSZ ) 

### Expected behavior

Implement --cluster= to filter for for returned results before further processing
List ALL Raft meta groups in the result. Making sure that RAFT meta groups are returned properly when --cluster= is applied

### Server and client version

Server 2.14 DEV
CLI 0.3.0

### Host environment

Windows

PR #1529, opened 2025-10-15
List multiple RAFT meta groups in server report jetstream 

roeschter:
Fix for https://github.com/nats-io/natscli/issues/1527

Observed behavior

When two cluster with diffrent domains are leaf node connected via a system account $SYS.REQ.SERVER.PING.JSZ will correctly return information from all nodes. AS cann be verified easily with --trace

The Jetstream summary will correctly list ALL nodes across two cluster but will only list ONE RAFT Meta Group Information.

This is somewhat annoying and confusing and makes it very hard to debug connected clusters with this command (short of looking at the raw data of $SYS.REQ.SERVER.PING.JSZ )
Expected behavior

List ALL Raft meta groups in the result.

### DIFF
--- a/cli/server_report_command.go
+++ b/cli/server_report_command.go
@@ -987,7 +987,7 @@ func (c *SrvReportCmd) reportJetStream(_ *fisk.ParseContext) error {
 			}
 
 			header := "RAFT Meta Group Information - Lead cluster: " + cluster.Name
-			table := iu.NewTableWriterf(opts(), header)
+			table := iu.NewTableWriterf(opts(), "%s", header)
 
 			table.AddHeaders("Connection Name", "ID", "Leader", "Current", "Online", "Active", "Lag")
 			for i, replica := range cluster.Replicas {


cli/server_report_command.go
```diff
@@ -956,52 +956,54 @@ func (c *SrvReportCmd) reportJetStream(_ *fisk.ParseContext) error {
...
+				online := "true"
+				if replica.Offline {
+					online = color.New(color.Bold).Sprint("false")
+				}
+
+				table.AddRow(cNames[i], peer, leader, replica.Current, online, f(replica.Active), f(replica.Lag))
 			}
 
-			table.AddRow(cNames[i], peer, leader, replica.Current, online, f(replica.Active), f(replica.Lag))
+			fmt.Print(table.Render())
```
ploubser: Might want to add a check here and print a newline if we are going to print multiple tables, just to give it some space on the screen.

cli/server_report_command.go
```diff
@@ -956,52 +956,54 @@ func (c *SrvReportCmd) reportJetStream(_ *fisk.ParseContext) error {
...
+			for _, r := range cluster.Replicas {
+				names = append(names, r.Name)
 			}
+			if c.compact {
+				cNames = iu.CompactStrings(names)
+			} else {
+				cNames = names
+			}
+
+			table := iu.NewTableWriterf(opts(), "RAFT Meta Group Information")
```
ploubser: Is it possible to distinguish between the tables here with an identifier in the Table heading?

roeschter: We can identify the cluster containing the leader.

PR #1549, opened 2025-11-03
Show details about how many messages were removed in purge

### LINKED ISSUE #1546: `nats stream purge` should show number of purged/removed messages
### Proposed change

Presently (as discussed in slack) `nats stream purge` doesn't show the amount of messages removed. The proposal is to clearly report the amount of messages deleted.

On a related note:

1. the additional report of "Deleted Messages" as output from purge isn't clear (relates to sequence gaps rather than messages purged)
2. the natscli help describes `purge` as `Purge a Stream without deleting it`, perhaps it would be more accurate now to describe it as `Purge a Stream or Stream Messages without deleting it`
3. possibly a --dry-run switch would be useful here as well?

### Use case

For applications using jetstream as a live store, rather than for buffering to provide at least once delivery, this is particularly useful for development and debugging. Showing the number of deleted messages provides immediate feedback to confirm expectation during development, and is useful wherever a filter is used to purge a subset of messages on the stream: presently --subject, --keep or a combination.

### Contribution

_No response_

ripienaar:
Also show state only not full info after purge

Fixes #1546


ploubser (approved):
👍

Issue #1498, opened 2025-09-19
Add cli version to audit gather metadata


alexbozhenko:
### Proposed change

When gather was part of the cli, we used to set version:
https://github.com/nats-io/natscli/commit/61c725c359dd3fb39b3d69680a2c98cf3a930d80#diff-452e58ada64db9f1034b213cadb3affe99b9bda948000988e1255dd007648146L691

we do not do not set any version anymore:
https://github.com/nats-io/jsm.go/blob/fa5c81909624897f9180f942a9ec8bcf8757b12a/audit/gather/gather.go#L273-L279

So we get empty version inside the archive:
```
# jq '.cli_version' audit_gather_metadata.json
""
```

Can we pass the version through this call as argument(or part of the config struct?)
https://github.com/nats-io/natscli/blob/5c1d0effcbda3be7a23b1ca8d4b32ec79398922b/cli/audit_gather_command.go#L62
So we can thread it to the metadata?

### Use case

We need to know version of the cli that was used to generated gather archive, e.g. when something does not look right in the output and we need to fiugre out why.

### Contribution

I can.

PR #1551, opened 2025-11-03
Record cli version when gathering, report in output

ripienaar:
Fixes https://github.com/nats-io/natscli/issues/1498


ploubser (approved):
👍

PR #1565, opened 2025-12-19
Audit gather - cluster filter

roeschter:
Allow to filter for a specific cluster. Keeping it simple. One cluster at a time or everything.

Reason: In large NATS meshes with many remote leaf clusters, as some of our customer use, it is not always desirable to get stats from all servers. In a globally distributed mesh it may take several minutes to run nats audit gather. While at the same time its desirable to retain the possibility to obtain all this info from a central location through connected SYS accounts.


roeschter:
Done

PR #1585, opened 2026-01-21
(maint) Separate pub and req commands

ploubser:
* Split pub and req into their own files
* Extract shared code into util
* Simplify and clarify pub logic


ploubser:
Holding off on this until after we release.

ripienaar (changes requested):

Issue #1593, opened 2026-02-04
nats req ignores --reply-timeout argument and always times out after 5 seconds


shadone:
### Observed behavior

I use request-reply and I cannot get the reply using nats-cli my server replies too slow - in my case it takes around 10 seconds to reply.

nats-cli req command has command line argument `--reply-timeout` but whatever I pass, it always times out after 5 seconds.

I follow this documentation https://docs.nats.io/nats-concepts/core-nats/reqreply/reqreply_walkthrough

This seems very similar to the other issue I reported previous - `nats obj get` was also timing out after 5 seconds https://github.com/nats-io/natscli/issues/1534

### Expected behavior

I expect nats-cli to wait long enough to get the reply from my server.

```sh
nats req --reply-timeout=50s hello hello
```

### Server and client version

```
$ nats --version
0.3.1
```

```
$ nats-server --version
nats-server: v2.12.4
```

### Host environment

macOS 15.6.1

```
$ uname -a
Darwin minbar.local 24.6.0 Darwin Kernel Version 24.6.0: Mon Jul 14 11:30:30 PDT 2025; root:xnu-11417.140.69~1/RELEASE_ARM64_T6020 arm64
```

### Steps to reproduce

Start reply client in one terminal

```sh
nats reply hello --command 'bash -c "sleep 10s && echo hi"'
```

Send a request from another terminal

```sh
# I expect this to timeout after 1 second but it still waits for 5 seconds instead
$ time nats req --reply-timeout=5s hello hello
19:03:34 Sending request on "hello"
nats req --reply-timeout=5s hello hello  0.02s user 0.02s system 0% cpu 5.036 total
```

```sh
# I expect this to get the reply after 10 seconds but instead it times out after 5 seconds
$ time nats req --reply-timeout=50s hello hello
19:03:18 Sending request on "hello"
nats req --reply-timeout=50s hello hello  0.01s user 0.01s system 0% cpu 5.022 total
```

PR #1594, opened 2026-02-11
(#1593) Update `--reply-timeout` doc string

ploubser:


cli/req_command.go
```diff
@@ -175,7 +175,7 @@ func (c *reqCmd) doReq(nc *nats.Conn, pub *iu.Publisher) error {
 			if c.replyCount == 0 {
 				// if we are waiting for the general timeout then
 				// calculate remaining
-				timeout = opts().Timeout - time.Since(start)
+				timeout = max(c.replyTimeout, opts().Timeout) - time.Since(start)
```
ripienaar: hmm, not sure this is right.

The idea here is a bit like a choria broadcast discover:

- you send the broadcast
- there could be no responses so you want to stop - this is on the opts().Timeout
- then when the flow of messages comes you dont want to end up with the weird 2 seconds delay at the end where it just sit and does nothing like Choria used to. Instead you count the time between most recent previous reply and compare that against the replyTimeout. replyTimeout signals to you the chunk of replies all came in and you are unlikely to see more unless things are broken

so we can't do max of the 2, the purpose here is specially to only accept the replyTimeout as maximum time spent between 2 replies.

I can show you how that works with Choria if you like

PR #1603, opened 2026-02-27
Populate missing Raft group in stream/consumer check

MauriceVanVeen:
Follow-up of https://github.com/nats-io/natscli/pull/1468. If the JSZ output doesn't contain full Raft data, the Raft group can still be populated from the Cluster info.


cli/server_consumer_check.go
```diff
@@ -149,6 +152,9 @@ func (c *ConsumerCheckCmd) consumerCheck(_ *fisk.ParseContext) error {
 							break
 						}
 					}
+					if raftGroup == "" {
+						raftGroup = consumer.Cluster.RaftGroup
```
ripienaar: I guess we need a nil guard here

PR #1613, opened 2026-03-13
Fix pub/req commands with piped stdin

MauriceVanVeen:
Tried piping stdin to the `nats pub` command but noticed this wasn't working anymore, and bisected this back to https://github.com/nats-io/natscli/pull/1585.

```sh
# Before this fix
echo "world" | nats pub hello
..
[#1] Received on "hello"
nil body

# After this fix
echo "world" | nats pub hello
..
[#3] Received on "hello"
world
```

The pub command was previously using `terminal.IsTerminal(int(os.Stdout.Fd())` but with the change to `isTerminal()` which also uses the stdin FD in its check this no longer evaluates to true: https://github.com/nats-io/natscli/pull/1585/changes#diff-61f207b144df96e68822452378155f0a5184f8d6994e0f2267e0e54f77faaf1bL433

### DIFF
--- a/internal/util/publisher.go
+++ b/internal/util/publisher.go
@@ -64,7 +64,7 @@ func NewPublisher(cfg PublisherConfig) (*Publisher, error) {
 		opts:      cfg.Opts,
 	}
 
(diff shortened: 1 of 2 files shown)


ploubser (approved):
👍

PR #1642, opened 2026-04-07
list all accounts in server report accounts command

zakk616:
- commnd `nats server report accounts` does not show all accounts untill they have active connections:

image

- the monitoring endpoint shows all accounts which is helpfull:

image

- fixed to show all accounts despit having no active connections:

image


ripienaar:
This should probably be opt-in as some times there can be many dormant accounts.

ripienaar:
Also be good to get tests in `TestServerReport`

ripienaar:
Closing this, review feedback was never addressed so its stale

PR #1647, opened 2026-04-15
feat: add --throughput flag to bench generator commands

jainal09:
## Summary
- Adds `--throughput THROUGHPUT` to `nats bench pub`, `bench service request`, `bench js pub sync|async|batch`, and `bench kv put`.
- When set `> 0`, throttles the aggregate publisher throughput to approximately `THROUGHPUT` messages/second across all clients. Each client targets `throughput / clients` msgs/sec via a simple expected-vs-elapsed sleep. `0` (default) disables throttling and leaves existing behavior unchanged.
- Mirrors the semantics of Kafka's `kafka-producer-perf-test.sh --throughput`, so NATS can be compared head-to-head under the same offered load rather than only at line rate.
- Docs: `README.md` benchmark section and `cli/cheats/bench.md` updated with an example.

## Why do we need this? (Kafka already has it)
Kafka's producer perf tool (`kafka-producer-perf-test.sh`) exposes `--throughput` precisely because benchmarking at line rate only tells you peak throughput. It does not tell you how the system behaves under a realistic, sustained offered load. Without an equivalent on `nats bench`, any NATS-vs-Kafka comparison is forced into one of two shapes:

1. Both run unthrottled: each system hits its own ceiling at a different rate, so you're comparing two different workloads. RAM/disk/CPU graphs are not directly comparable.
2. Rate-limit Kafka but let NATS free-run: NATS looks artificially better on latency but is doing 10–100× more work, which invalidates the comparison entirely.

**A throughput cap fixes this.** When you hold the offered message rate constant across both systems, you can actually compare the internals that matter in a sustained-load scenario:

- **RAM usage over time**: does the broker's memory drift up as backlog or buffers grow?
- **Disk usage / write amplification**: at the same 50k msgs/sec, how much do on-disk segments grow, and how fast? How big are the WAL/stream files after N minutes?
- **CPU utilization at matched load**: who spends more cycles per delivered message?
- **Latency distributions (p50/p99/p99.9) at a fixed rate**: this is the number operators actually care about, and it's only meaningful at matched input rates.
- **GC pressure, fsync behavior, replication lag**: all of these only make sense to compare when both brokers are being asked to do the same amount of work.

In short: line-rate benchmarks answer "how fast can it go?" Throttled benchmarks answer "what does it cost to run at rate X?". The second question is the one that drives capacity planning and head-to-head architecture decisions. This flag enables the second kind of test for NATS.

## Implementation notes
- New `rateThrottler` (expected = sent / target sec; sleeps the delta if ahead) created per publisher goroutine in `cli/bench_command.go`.
- Wired into `coreNATSPublisher`, `coreNATSRequester`, `jsPublisher` (sync / async / batch branches), and `kvPutter`.
- Banner prints `throughput=N` only for the generator-side bench types when the flag is set.

## Test plan
Tested locally against a `nats-server -js` on port `14222`:

- [x] Unthrottled baseline: `bench pub` → 5,183,108 msgs/sec
- [x] `bench pub --throughput 5000` (1 client, 10k msgs) → **4,999 msgs/sec**
- [x] `bench pub --throughput 20000 --clients 4` (40k msgs) → aggregate **19,986 msgs/sec**, per-client ~4,996
- [x] `bench js pub sync --throughput 2500` → **2,499 msgs/sec**
- [x] `bench kv put --throughput 1000` → **999 msgs/sec**
- [x] `go build ./cli/... ./internal/...` clean
- [x] `go vet ./cli/...` clean
- [x] `nats bench pub --help` shows the new flag

Backwards compatible: omitting the flag (or passing `0`) preserves current behavior exactly.


jainal09:
Addressed all three points in 30d6e63:

1. **JS async per-batch bursts** — moved the throttle call inside the inner `PublishMsgAsync` loop, pacing on attempted sends (`i+j+1`) rather than once after the whole batch + ack cycle. A batch can no longer fire at line rate before a catch-up sleep.
2. **JS batch per-batch bursts** — same fix applied to the `TypeJSPubBatch` inner loop (including the commit message), so send rate stays near the configured target across a batch's lifetime.
3. **Hot-loop overhead when disabled** — `newRateThrottler` now returns `nil` when the target is non-positive, and every call site guards with `if throttler != nil { ... }`. The `--throughput=0` path has no added work beyond a single nil compare per iteration.

Retested locally against `nats-server -js`:

- Unthrottled baseline: `bench pub` → **5,012,147 msgs/sec** (no regression vs main)
- `bench js pub async --batch 500 --throughput 500` (the bursty scenario flagged in #1) → **499 msgs/sec**, per-batch wall time ~1,000,075µs, i.e. each batch now takes `batch_size / rate` seconds to emit rather than firing instantly + sleeping
- `bench js pub batch --batch 500 --throughput 500` → **499 msgs/sec**, same 1s/batch pacing
- Re-ran the original sanity case: `bench pub --throughput 5000` → **4,998 msgs/sec**

jainal09:
Copilot second pass. Addressed all three new comments in 95a629a:

1. **JS async retry accounting (L2011)** — good catch. Since `i` only advances on successful acks, `i+j+1` under-counts attempted publishes on retries and the throttler would sleep less than it should. Introduced a dedicated `attempts` counter incremented on every `PublishMsgAsync` call, and pacing is now driven off that. On the happy (no-error) path `attempts == i+j+1` so there is no behavior change; on ack failures the throttler now correctly sees the retry attempts.
2. **Flag help wording (L146)** — applied your suggestion almost verbatim. Now reads "aggregate message send throughput" and notes the `--sleep` interaction.
3. **README `--sleep` composition (L489)** — applied your suggestion verbatim, so readers aren't surprised when `--sleep` caps the achieved rate below `--throughput`.

Re-ran the async burst scenario after the `attempts` change: `bench js pub async --batch 500 --throughput 500` → **499 msgs/sec**, pacing still ~1s per 500-msg batch as before. No regression on the happy path.

jainal09:
@jnmoyne  or @ripienaar  does the tests needs approval of some kind to run ? As it says `1 workflow awaiting approval`

Could you please run the workflow for me? In order for me to fix any breaking stuff in my PR

jainal09:
Hi, could you please approve the latest CI run? The previous run failed on `TestConsumerClusterDown` with `no suitable peers for placement, peer offline (10005)`, which is a flaky cluster-setup race unrelated to this PR. The same test passed on 3 of the 4 test matrix entries (ubuntu/1.26, windows/1.25, windows/1.26) and only failed on ubuntu/1.25. This PR only touches `cli/bench_command.go`, `cli/cheats/bench.md`, and `README.md`, no overlap with consumer command code. Pushed an empty commit to trigger a fresh run. Thanks!

jainal09:
Great all CI runs passes!
LMK if you want any changes. Or else this PR is ready to be merged!

jnmoyne:
Hi and thanks for the contribution! Could you please sign your commits, and have a 'signed-off-by'? (you could for example squash all the commits down to one, that is signed and have the signed-off line in the description)

jainal09:
@jnmoyne  Done! Squashed all commits into a single signed commit with a Signed-off-by line. Thanks for the
  heads-up!

jnmoyne (approved):
LGTM

Issue #1645, opened 2026-04-14
Report issystem account in account info


ripienaar:
Since https://github.com/nats-io/nats.go/releases/tag/v1.51.0 there is now a way to know if you are using a system account, expose that in `nats account info`

PR #1648, opened 2026-04-15
(#1645) - Report issystem account in account info

ploubser:
* Add `System Account` to nats account info output
* Add tests for account commands
* Bump nats.go dependency


cli/account_command.go
ripienaar, line 467: Maybe move it up to here so its together with other account name etc

ripienaar (approved):

PR #1663, opened 2026-05-11
doc enhancements

roeschter:
Clarification on documentation.
1) flow control option is ambiguous
2) source and mirror for stream create accept JSON (those are the only parameter which behave like this)

### DIFF
--- a/cli/consumer_command.go
+++ b/cli/consumer_command.go
@@ -154,7 +154,7 @@ func configureConsumerCommand(app commandHost) {
 		}
 		f.Flag("filter", "Filter Stream by subjects").PlaceHolder("SUBJECTS").StringsVar(&c.filterSubjects)
 		if !edit {
-			f.Flag("flow-control", "Enable Push consumer flow control").IsSetByUser(&c.fcSet).UnNegatableBoolVar(&c.fc)
+			f.Flag("flow-control", "Enable Push consumer flow control. Note: This is distinct from the 'flow_control` ack mode.").IsSetByUser(&c.fcSet).UnNegatableBoolVar(&c.fc)
 			f.Flag("heartbeat", "Enable idle Push consumer heartbeats (-1 disable)").StringVar(&c.idleHeartbeat)
 		}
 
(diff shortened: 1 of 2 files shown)


cli/consumer_command.go
```diff
@@ -154,7 +154,7 @@ func configureConsumerCommand(app commandHost) {
 		}
 		f.Flag("filter", "Filter Stream by subjects").PlaceHolder("SUBJECTS").StringsVar(&c.filterSubjects)
 		if !edit {
-			f.Flag("flow-control", "Enable Push consumer flow control").IsSetByUser(&c.fcSet).UnNegatableBoolVar(&c.fc)
+			f.Flag("flow-control", "Enable Push consumer flow control. Note: This is distinct from the 'flow_control` ack mode.").IsSetByUser(&c.fcSet).UnNegatableBoolVar(&c.fc)
```
ripienaar, line 157: we dont add notes like this in flag help, this would be the only example like this, these flags help is quick glance help and in the case of something so complex and so full of many 10s of flags the expectation is you're following documentation when using these esoteric options.

That's also why they aren't requested during the interactive question/answer phase.

Too long and verbose

cli/stream_command.go
```diff
@@ -222,11 +222,11 @@ func configureStreamCommand(app commandHost) {
 		f.Flag("max-msgs", "Maximum amount of messages to keep").Default("0").Int64Var(&c.maxMsgLimit)
 		f.Flag("max-msgs-per-subject", "Maximum amount of messages to keep per subject").Default("0").Int64Var(&c.maxMsgPerSubjectLimit)
 		f.Flag("dupe-window", "Duration of the duplicate message tracking window").Default("").StringVar(&c.dupeWindow)
-		f.Flag("mirror", "Completely mirror another stream").StringVar(&c.mirror)
+		f.Flag("mirror", "Completely mirror another stream. Accepts mirror config as JSON. ").StringVar(&c.mirror)
```
ripienaar, line 225: This implies this all it accepts :) Just say like `Completely mirror another stream, stream name or configuration JSON`

ripienaar (changes requested):

PR #1665, opened 2026-05-12
Rework the 'nats context' command to use registry

ripienaar:
This is step one before looking into supporting pluggable backends so for now just keeping like for like behavior except for one case of error recovery in edit which we marked for a later phase todo

### DIFF
--- a/cli/context_command.go
+++ b/cli/context_command.go
@@ -15,12 +15,13 @@ package cli
 
 import (
 	"bytes"
-	"encoding/json"
+	"errors"
 	"fmt"
 	"os"
 	"os/user"
 	"sort"
 	"strings"
+	"sync"
 	"text/template"
 
 	iu "github.com/nats-io/natscli/internal/util"
@@ -44,6 +45,9 @@ type ctxCommand struct {
 	force            bool
 	validateErrors   int
 	embed            bool
+
+	reg *natscontext.Registry
+	mu  sync.Mutex
 }
 
 func configureCtxCommand(app commandHost) {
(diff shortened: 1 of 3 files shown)


go.mod
```diff
@@ -62,7 +62,7 @@ require (
 	github.com/mitchellh/go-homedir v1.1.0 // indirect
 	github.com/mitchellh/reflectwalk v1.0.2 // indirect
 	github.com/munnerz/goautoneg v0.0.0-20191010083416-a7dc8b61c822 // indirect
-	github.com/nats-io/nsc/v2 v2.12.2 // indirect
+	github.com/nats-io/nsc/v2 v2.10.3-0.20250110165315-eeda721ecff6 // indirect
```
ploubser: This on purpose?

ripienaar: no harm, we fix these things when we release

ripienaar: but let me see this is weird

ploubser (approved):

cli/context_command.go
```diff
@@ -109,6 +113,29 @@ func init() {
...
+	c.mu.Lock()
+	defer c.mu.Unlock()
+
+	if len(opts) == 0 && c.reg != nil {
+		return c.reg
+	}
+
+	registryOpts := []natscontext.RegistryOption{
+		natscontext.WithDefaultResolvers(),
+		natscontext.WithLocalSelector(),
```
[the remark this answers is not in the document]
ripienaar: fine for now, this is a step in a process to moving to the new registry based

PR #1672, opened 2026-05-27
fix: use `nats.ProxyPath` when server is a websocket connection with a path

joeriddles:
Fix WebSocket connections when the URL has a path.

This is an alternative fix to https://github.com/nats-io/nats.go/pull/2092.

```shell
# before
❯ nats -s 'ws://localhost:8080/api/cp-nats/' pub 'xyz' test
nats: error: invalid websocket connection

# after
❯ nats -s 'ws://localhost:8080/api/cp-nats/' pub 'xyz' test
10:22:37 Published 4 bytes to "xyz"
````

### DIFF
--- a/cli/util.go
+++ b/cli/util.go
@@ -400,19 +400,20 @@ func newNatsConnUnlocked(servers string, copts ...nats.Option) (*nats.Conn, erro
 		servers = opts.Config.ServerURL()
 	}
 
-	servers, proxyPath := extractWSProxyPath(servers)
+	servers, proxyPath, err := extractWSProxyPath(servers)
+	if err != nil {
+		return nil, err
+	}
 	if proxyPath != "" {
 		copts = append(copts, nats.ProxyPath(proxyPath))
 	}
-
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: cli/util.go [hunk 1]
```diff
@@ -394,13 +400,54 @@ func newNatsConnUnlocked(servers string, copts ...nats.Option) (*nats.Conn, erro
 		servers = opts.Config.ServerURL()
 	}
 
-	var err error
+	servers, proxyPath, err := extractWSProxyPath(servers)
+	if err != nil {
+		return nil, err
+	}
+	if proxyPath != "" {
+		copts = append(copts, nats.ProxyPath(proxyPath))
```
ripienaar, line 408: this would be a natscontext behaviour instead wouldn't it? It handles the core connection ops setting and parsing etc

joeriddles: I created a PR in jsm.go: https://github.com/nats-io/jsm.go/pull/813

PR #1674, opened 2026-05-28
Add --tls-first as a hidden alias for --tlsfirst

### LINKED ISSUE #1673: Add --tls-first alias to match nsc's flag spelling
### Proposed change

Register `--tls-first` as an alias for `--tlsfirst` so either spelling works. No behavior change, both names bind to the same `opts.TlsFirst`. fisk has `.Alias()` for exactly this case.

Around [`nats/main.go:78`](https://github.com/nats-io/natscli/blob/main/nats/main.go#L78):

```go
ncli.Flag("tlsfirst", "Perform TLS handshake before expecting the server greeting").
    Alias("tls-first").
    BoolVar(&opts.TlsFirst)
```

Happy to flip the canonical direction (alias `--tlsfirst`, canonical `--tls-first`) if maintainers prefer that — the goal is just for either spelling to work in either tool.

### Use case

The `nats` CLI exposes the TLS-first connection option as `--tlsfirst` (one word). The [nsc CLI](https://github.com/nats-io/nsc) exposes the same option as `--tls-first` (hyphenated). Both wrap the same underlying `nats.go` `TLSHandshakeFirst`, but the inconsistent spelling trips up anyone scripting against both tools.

Concrete reproducer — both commands target the same Istio-fronted NATS deployment that requires TLS-first:

```bash
nats --tlsfirst  --server tls://... --creds sys.creds rtt        # works
nsc  --tls-first --server nats://... --account weather push      # works

# Run either with the other's flag:
nats --tls-first  ...  →  nats: error: unknown long flag '--tls-first'
nsc  --tlsfirst   ...  →  nsc:  Error: unknown flag: --tlsfirst
```

Each tool's `--help` shows only the local spelling, so users discover the inconsistency by trial and error in CI. Common admin/CI pattern is to use both binaries (`nsc` for identity management, `nats` for runtime/$SYS queries) where the asymmetry bites repeatedly.

Filing the matching request against [nsc](https://github.com/nats-io/nsc/issues/824) so both CLIs accept either
spelling.

### Contribution

Yes, happy to open the PR. The change is a single-line `.Alias("tls-first")` addition on the existing `BoolVar(&opts.TlsFirst)` chain. Will follow up here once the direction's confirmed.

pjaudiomv:
Closes #1673.

### DIFF
--- a/nats/main.go
+++ b/nats/main.go
@@ -76,6 +76,10 @@ LLM optimized help output can be obtained using --help-llm for any command. You 
 	ncli.Flag("tlskey", "TLS private key").Envar("NATS_KEY").PlaceHolder("FILE").ExistingFileVar(&opts.TlsKey)
 	ncli.Flag("tlsca", "TLS certificate authority chain").Envar("NATS_CA").PlaceHolder("FILE").ExistingFileVar(&opts.TlsCA)
 	ncli.Flag("tlsfirst", "Perform TLS handshake before expecting the server greeting").BoolVar(&opts.TlsFirst)
+	// Alias to match nsc's spelling (--tls-first). Hidden so --help keeps
+	// showing only the canonical --tlsfirst; both names bind to the same
+	// variable so either spelling works in scripts.
+	ncli.Flag("tls-first", "alias for --tlsfirst").Hidden().BoolVar(&opts.TlsFirst)
 	if runtime.GOOS == "windows" {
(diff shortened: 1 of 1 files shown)


ripienaar:
Thank you @pjaudiomv

pjaudiomv:
Thanks 🙏

PR #1679, opened 2026-06-04
README: How to add authentication

holyjak:
Context:
The readme wasn't providing any information about how to authenticate with the NATS server, and it wasn't trivial to figure this out.

Action:
Mention that authentication config can be added to the context.


ripienaar:
Thanks, to be honest I should trim the readme down to a link to the official docs site.

PR #1682, opened 2026-06-15
add websocket headers and proxy path support

ktarplee:
This PR adds CLI flags for websocket headers and proxy path.

Some thoughts:
1. While the NATS API separates the proxy path from the server name, is that what we would want in the API?  I think it would be cleaner to support a server URL and then extract the URL path from that and use it as the proxy path.  `-s wss://example.com/path/to/websocket` as opposed to `-s wss://example.com --proxy-path /path/to/websocket`.
2. The NATS context config file should probably support setting the websocket headers and proxy path but that object seems to be defined in a the jsm.js project for some reason.   It is not obvious to me why the config file for the NATS cli is in the jsm.go project.
3. I still need to fix the `rtt` command.  It does some stuff differently that does not seem to work with websockets.

### DIFF
--- a/cli/util.go
+++ b/cli/util.go
@@ -19,6 +19,7 @@ import (
 	"fmt"
 	"io"
 	"math"
+	"net/http"
 	"os"
 	"os/exec"
 	"sort"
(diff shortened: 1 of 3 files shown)


ripienaar:
We have other inflight PRs around this here https://github.com/nats-io/jsm.go/pull/813 and here https://github.com/nats-io/natscli/pull/1672#discussion_r3313123178 and https://github.com/nats-io/nats.go/pull/2092

I have no idea what the correct approach is but seems there are some contradictory approaches here

Issue #1683, opened 2026-06-15
Add a leafnode connection count to server ls


jarretlavallee:
### Proposed change

Add a leafnode connection count to `nats server ls` (e.g. a `Leafs` column).

### Use case

The number of leafnode connections on a server can affect its performance. You can rebalance them with `nats server cluster balance --kind Leafnode`, but it's hard to know when a server actually needs it. `nats server report leafnodes` lists every connection but doesn't total them per server, and `nats server ls` doesn't show leaf counts.

It's also useful for sizing up an environment. `nats server ls` already gives a good overview of the cluster, but leafnodes are a blind spot, so you can't see how big the deployment really is once leaf connections are counted.

A leaf count in `nats server ls` would let you spot an imbalance, know when to run balance, and get a true picture of the environment size.

PR #1692, opened 2026-07-01
(#1683) Add leafnode connection count to server ls

ploubser:


ploubser:
@codex review

cli/server_list_command.go
```diff
@@ -144,6 +157,20 @@ func (c *SrvLsCmd) list(_ *fisk.ParseContext) error {
 			defer mu.Unlock()
 			accumulate(ssm, time.Since(start))
 		})
+
+		serverdata.DoReqAsync(ctx, nil, "$SYS.REQ.SERVER.PING.VARZ", int(c.expect), nc, opts().Timeout, traceLogger(), func(data []byte) {
```
ripienaar, line 161: hmm, not that keen on introducing another hop here, maybe best to park this one and get ServerStatsMsg to include leaf counts?

ploubser: Yeah that's probably a good idea. Adding another hop is also probably going to mess with TTL's. Will park and investigate.

ripienaar: Should be easy to contribute that to the server (foreshadowing no doubt)

Issue #1688, opened 2026-06-29
server cluster peer-remove crashes when there is more than one meta raft visible (connected system accounts)


roeschter:
### Observed behavior

Somewhat annoying crash in the CLI. Happens in main branch (v0.4.1-0.20260624121635-885fc11ab694)  as well as older versions.

Seems to be triggered by more than one Meta Raft group visible (here is leaf cluster connected to a central supercluster)

- --force does NOT bypass it
- --CLI 3.2 already has the issue
- --Issue happens during info/validation - NOT while sending the actuall request

**Tested with supercluster on itself - no issues**

**Workaround is  to send request directly:**
nats --server=127.0.0.1:7201 --user=admin --password=admin request '$JS.API.SERVER.REMOVE' '{"peer":"C1","peer_id":""}'

D:\synadia\software\nats\nats.exe --server=127.0.0.1:7201 --user=admin --password=admin server cluster peer-remove C3

```

    panic: runtime error: invalid memory address or nil pointer dereference
[signal 0xc0000005 code=0x0 addr=0x30 pc=0x7ff7353e000c]

goroutine 1 [running]:
github.com/nats-io/natscli/cli.(*SrvClusterCmd).metaPeerRemoveAction(0xc0000de3c0, 0x7ff7349677df?)
        D:/private/nats.io_synadia/natscli/cli/server_cluster_command.go:201 +0xec
github.com/choria-io/fisk.(*actionMixin).applyActions(...)
        C:/Users/micha/go/pkg/mod/github.com/choria-io/fisk@v0.8.4-0.20260530151536-30471af3a46f/actions.go:28
github.com/choria-io/fisk.(*Application).applyActions(0xc0002ad340?, 0xc00038a1b0)
        C:/Users/micha/go/pkg/mod/github.com/choria-io/fisk@v0.8.4-0.20260530151536-30471af3a46f/app.go:842 +0xd0
github.com/choria-io/fisk.(*Application).execute(0xc0002ad340, 0xc00038a1b0, {0xc0002717c0, 0x3, 0x4})
        C:/Users/micha/go/pkg/mod/github.com/choria-io/fisk@v0.8.4-0.20260530151536-30471af3a46f/app.go:643 +0x65
github.com/choria-io/fisk.(*Application).Parse(0xc0002ad340, {0xc000050090?, 0x2c?, 0x7ff736299f80?})
[...]

PR #1693, opened 2026-07-01
(#1688) Fix panic in peer-remove subcommand

ploubser:
Skip JSZ responses without a meta group instead of dereferencing a nil
Meta which caused a panic.

If multiple meta leaders are detected we will now exit with an
explicit message instead of assuming which one to remove the peer from.

### DIFF
--- a/cli/server_cluster_command.go
+++ b/cli/server_cluster_command.go
@@ -174,7 +174,7 @@ func (c *SrvClusterCmd) metaPeerRemoveAction(_ *fisk.ParseContext) error {
 	reqFn := func(req any, subj string, waitFor int, nc *nats.Conn) ([][]byte, error) {
 		return serverdata.DoReq(ctx, req, subj, waitFor, nc, opts().Timeout, traceLogger())
 	}
-	ds, err := serverdata.NewLive(nc, reqFn, 1)
+	ds, err := serverdata.NewLive(nc, reqFn, -1)
 	if err != nil {
 		return err
 	}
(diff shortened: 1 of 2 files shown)


cli/server_cluster_command.go
```diff
@@ -184,28 +184,46 @@ func (c *SrvClusterCmd) metaPeerRemoveAction(_ *fisk.ParseContext) error {
 		return err
 	}
 
-	if len(jszResults) != 1 {
+	if len(jszResults) == 0 {
 		return fmt.Errorf("did not receive a response from the meta leader, ensure the account used has system privileges and appropriate permissions")
 	}
 
+	targetDomain := opts().Config.JSDomain()
```
ripienaar, line 191: think we need some comments here to describe what this is about

cli/server_cluster_command.go
```diff
@@ -184,28 +184,46 @@ func (c *SrvClusterCmd) metaPeerRemoveAction(_ *fisk.ParseContext) error {
...
+	if targetDomain == "" {
+		connectedJsz, err := ds.Jsz(server.JszEventOptions{EventFilterOptions: server.EventFilterOptions{Name: nc.ConnectedServerName(), ExactMatch: true}})
```
ripienaar: since you set the filter...

cli/server_cluster_command.go
```diff
@@ -184,28 +184,46 @@ func (c *SrvClusterCmd) metaPeerRemoveAction(_ *fisk.ParseContext) error {
...
+		if err != nil {
+			return err
+		}
+		for _, srv := range connectedJsz {
+			if srv.Data != nil && srv.Data.ID == nc.ConnectedServerId() {
```
ripienaar: you should never get one where this condition is false?

Seems a upfront call to `$SYS.REQ.SERVER.PING.VARZ` when your domain isn't set (filtered to connected server id) should probably do this quicker and easier?

ripienaar (changes requested):

ploubser:
@codex review

ploubser:
@codex review

ploubser:
@ripienaar what's your preference with the timeout parameter? 0 or -1? the bot keeps alternating it's findings 🤣

ripienaar:
We need to figure out if there are multiple domains so I think a long timeout is right and if anything at all responds from a different domain or we get 2 leaders then we error

ploubser:
@ripienaar agree. What we currently have here is imo the most defensive approach. Going to resolve the codex comment because it's going to flip to the other thing no matter what we do.

ploubser:
Going to close this PR and open a new one. The work here has drifted beyond just trying to fix a panic into trying to determine the correct behaviour for peer removal as a thing.

PR #1694, opened 2026-07-02
(#1688) Don't allow peer-removal when we find multiple leaders

ploubser:
* Because peer-remove and step-down uses the system account, we are unable to
prefix the call with a specific domain. This causes unexpected errors at best
and removal of the wrong peer at worst. We update the logic to only allow peer
removal when we detect a single leader for now.

* Also skip JSZ responses without a meta group instead of dereferencing a nil
Meta which caused a panic.

* Change the way we process leaders. Instead of doing a leader_only ping with
a timeout we get all the servers and then wait for each of them to respond.
This makes the common case (single cluster) fast but also lets us get all hosts
before making destructive calls.

Signed-off-by: Pieter Loubser <ploubser@gmail.com>

### DIFF
--- a/cli/server_account_command.go
+++ b/cli/server_account_command.go
@@ -43,6 +43,14 @@ func configureServerAccountCommand(srv *fisk.CmdClause) {
 }
 
 func (c *srvAccountCommand) purgeAccount(_ *fisk.ParseContext) error {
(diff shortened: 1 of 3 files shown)


ploubser:
@codex review

ploubser:
@codex review

ripienaar (approved):

PR #1705, opened 2026-08-03
Filter pending bytes

roeschter:
A (no  longer) fairly simple change that will make looking for slow consumers at lot easier.
Update nats cheat request

Example:
`nats server req connections --subscriptions --filter-empty --limit=64000 --filter-pending-bytes=48000000`
OR
nats server req connections --subscriptions --filter-empty --limit=64000 --filter-pending-bytes=75%

Also supported in:
`nats server report connections --pending-bytes=48000000`
`nats account report connections --pending-bytes=48000000`

### DIFF
--- a/cli/account_command.go
+++ b/cli/account_command.go
@@ -54,6 +54,7 @@ type actCmd struct {
 	stateFilter      string
 	user             string
 	filterReason     string
+	pendingBytes     string
 }
 
 func configureActCommand(app commandHost) {
@@ -67,7 +68,7 @@ func configureActCommand(app commandHost) {
 	report := act.Command("report", "Report on account metrics").Alias("rep")
 	report.Tag("scope:user", "impact:ro")
 
-	conns := report.Command("connections", "Report on connections").Alias("conn").Alias("connz").Alias("conns").Action(c.reportConnectionsAction)
+	conns := report.Command("connections", "Report on connections. For more options try 'nats server req connections'").Alias("conn").Alias("connz").Alias("conns").Action(c.reportConnectionsAction)
 	conns.Tag("scope:user", "impact:ro")
 	conns.Flag("sort", "Sort by a specific property (in-bytes,out-bytes,in-msgs,out-msgs,uptime,cid,subs)").Default("subs").EnumVar(&c.sort, "in-bytes", "out-bytes", "in-msgs", "out-msgs", "uptime", "cid", "subs")
 	conns.Flag("top", "Limit results to the top results").Default("1000").IntVar(&c.topk)
@@ -76,6 +77,7 @@ func configureActCommand(app commandHost) {
 	conns.Flag("reverse", "Reverse sort connections").Short('R').UnNegatableBoolVar(&c.reverse)
 	conns.Flag("state", "Limits responses only to those connections that are in a specific state (open, closed, all)").Default("open").EnumVar(&c.stateFilter, "open", "closed", "all")
 	conns.Flag("closed-reason", "Filter results based on a closed reason").PlaceHolder("REASON").StringVar(&c.filterReason)
+	conns.Flag("pending-bytes", "Limits responses only to those connections with at least this many pending bytes, supports sizes like 1MB").PlaceHolder("BYTES").StringVar(&c.pendingBytes)
 
 	stats := report.Command("statistics", "Report on server statistics").Alias("stats").Alias("statsz").Action(c.reportServerStats)
 	stats.Tag("scope:user", "impact:ro")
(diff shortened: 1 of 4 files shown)


cli/server_request_command.go
```diff
@@ -587,6 +589,18 @@ func (c *SrvRequestCmd) conns(_ *fisk.ParseContext) error {
...
+		if c.pendingBytesFilter > 0 && r.Data != nil {
+			matched := make([]*server.ConnInfo, 0, len(r.Data.Conns))
+			for _, conn := range r.Data.Conns {
+				if conn != nil && conn.Pending >= c.pendingBytesFilter {
+					matched = append(matched, conn)
+				}
+			}
+			r.Data.Conns = matched
+			r.Data.NumConns = len(matched)
+			r.Data.Total = len(matched)
```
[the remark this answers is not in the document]
roeschter: OK - added example in nats cheat request

bruth:
Another option could be to make this percentage-based relative to the server-configured max pending limit.

roeschter:
> server

Although this requires system priviliges

ripienaar:
I can look a bit later - there is also a account scoped version of this command though 

https://github.com/nats-io/natscli/blob/c778a5c73818fc74d36f8504f2290a4ae9ebf561/cli/account_command.go#L70

roeschter:
> I can look a bit later - there is also a account scoped version of this command though
> 
> https://github.com/nats-io/natscli/blob/c778a5c73818fc74d36f8504f2290a4ae9ebf561/cli/account_command.go#L70

Thats just an alias (I tried it works account scoped) but I may need to update the docs.

ripienaar:
It’s not just an alias. You have to add the flag and pass in tbe option.

roeschter:
> It’s not just an alias. You have to add the flag and pass in the option.

Right. Lets see how we can make this work. 
nats account report connections is an alias to nats server report connections 
added the option in both

cli/account_command.go
```diff
@@ -67,7 +68,7 @@ func configureActCommand(app commandHost) {
 	report := act.Command("report", "Report on account metrics").Alias("rep")
 	report.Tag("scope:user", "impact:ro")
 
-	conns := report.Command("connections", "Report on connections").Alias("conn").Alias("connz").Alias("conns").Action(c.reportConnectionsAction)
+	conns := report.Command("connections", "Report on connections. For more options try 'nats server req connections'").Alias("conn").Alias("connz").Alias("conns").Action(c.reportConnectionsAction)
```
ripienaar, line 71: please revert this, as a end user command they can't do anything with this advise they aren't system users

cli/server_report_command.go
```diff
@@ -1391,6 +1393,25 @@ func (c *SrvReportCmd) getConnz(limit int, nc *nats.Conn) (connzList, error) {
 		fisk.FatalIfError(err, "Invalid expression: %v", err)
 	}
 
+	pendingBytes, pendingPercent, err := parsePendingBytesFilter("--pending-bytes", c.pendingBytesFilter)
+	if err != nil {
+		return nil, err
+	}
+	if pendingPercent > 0 {
+		return nil, fmt.Errorf("--pending-bytes does not support percentages, use 'nats server request connections --filter-pending-bytes' instead")
```
ripienaar, line 1401: why do we need different flags? Seems to me one flag can support both types of filter

cli/server_request_command.go
```diff
@@ -549,6 +554,58 @@ func (c *SrvRequestCmd) routez(_ *fisk.ParseContext) error {
 	return printResults(responses)
 }
 
+// parsePendingBytesFilter parses a pending bytes filter which is either an absolute
+// size like 1024 or 1MB, or a percentage like 10%. Exactly one of the returned values
+// will be non zero, both are 0 when the filter is not set.
+func parsePendingBytesFilter(flag string, val string) (int64, float64, error) {
```
ripienaar, line 560: these kinds of functions go in `internal/util` so we dont make a mess of the cli package, and they need tests then

cli/server_request_command.go
```diff
@@ -581,12 +638,51 @@ func (c *SrvRequestCmd) conns(_ *fisk.ParseContext) error {
 		opts.State = server.ConnOpen
 	}
 
+	pendingBytes, pendingPercent, err := parsePendingBytesFilter("--filter-pending-bytes", c.pendingBytesFilter)
```
ripienaar, line 641: `nats server req` is by design equivalent of the http endpoints, the only filtering we support here is the ones the http and requests support - ie. server side filtering.

So we should remove all the filtering from this command, it's not in line with the overall design or purpose.  `report` is for custom stuff

ripienaar (changes requested):
some comments, lets address those and I'll take another closer look

cli/server_request_command.go
roeschter: OK - This means a major revamping of nats server report connections - as we need subscription details to make this be useful.

roeschter:
Closing this PR and starting over

ripienaar:
You don’t need server changes you can add these features in this way to report commands - it’s just “server req” is just a equivalent to what’s available under http

but you can absolutely expand server report for these features we have lots of similar expansions there already

it’s for the same reason that report commands don’t have json output as they are not standard - but req does as it’s purely standard behaviours. 

So I am not saying don’t do this just focus on tbe report side and implement there

PR #1713, opened 2026-08-27
Add --connect-timeout option

philpennock:
The existing --timeout only governs how long to wait on NATS protocol layer responses, after the connection has been established.  The connection setup itself has always used the nats.go default of 2s.

Add a --connect-timeout flag ($NATS_CONNECT_TIMEOUT) to set nats.Timeout, for the connection flow before NATS protocol messages start to be exchanged.

_(Implemented by Claude Fable at my direction, I have reviewed the generated code and am happy with it.)_

### DIFF
--- a/cli/context_command.go
+++ b/cli/context_command.go
@@ -562,12 +562,15 @@ func (c *ctxCommand) showCommand(_ *fisk.ParseContext) error {
 	cols.AddRowIfNotEmpty("Color Scheme", cfg.ColorScheme())
 
 	checkConn := func() error {
-		opts, err := cfg.NATSOptions()
-		opts = append(opts, nats.MaxReconnects(1))
+		nopts, err := cfg.NATSOptions()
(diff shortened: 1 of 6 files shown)


ripienaar:
I dont think including links to Claude session output is appropriate for commit messages, please remove.

Will review in detail late this week

ripienaar:
The commit is also not Signed Off. Please just rewrite this commit message with your own words and brain @philpennock

ripienaar (changes requested):
Change is fine but:

 - Please rewrite commit and PR body adding claude session output isn't appropriate
 - Commits are required to be signed off

Best to just write these in your own words.

philpennock:
Okay, I did the PR text entirely in my own words but had let Claude do the commit.  Will address.

philpennock:
Converted to draft while examining scope of other potential connections which might have been missed before.

Issue #1714, opened 2026-09-01
Flag `--raw` causes `{{Subject}}` to expand to an empty string when used with `--translate`


axjab:
### Observed behavior

|natscli version|
|-|
|0.4.0|

I am trying to remove the `[#1] Received on "foo"` output for a cleaner, more compact display. As far as I can tell, the only way to do this is by using the `--raw` flag. However, doing so causes the `{{Subject}}` template variable in `--translate` to expand to an empty string, so the subject is no longer available to the translator.

---

**Without** `--raw`:
```bash
> nats sub --translate "sed 's/^/{{Subject}}: /'" ">"
Subscribing on > 
[#1] Received on "foo"
foo: bar
```

**With** `--raw`:
```bash
> nats sub --raw --translate "sed 's/^/{{Subject}}: /'" ">"
: bar
```

### Expected behavior

**With** `--raw`:
```bash
> nats sub --raw --translate "sed 's/^/{{Subject}}: /'" ">"
foo: bar
```

### Server and client version

**nats-server**: 2.14.6
**nats**: 0.4.0

### Steps to reproduce

**Terminal 1**:
```
nats sub --raw --translate "sed 's/^/{{Subject}}: /'" ">"
```

**Terminal 2**:
```
nats pub foo bar
```

**Expected**: `foo: bar`
**Received**: `: bar`

PR #1716, opened 2026-09-03
(#1714) Pass Subject when using --raw flag

ploubser:


ploubser:
Addressed comments, @codex review

ploubser:
@codex review

ploubser:
The bot's complaint make sense, but this is how --translate has always worked for every caller. This probably isn't the place to change it, since fixing a bug is turning into a discussion about how values reach the command. 

@codex review

ripienaar (approved):

Issue #1676, opened 2026-06-01
Support editing Server 2.15 backups


ripienaar:
### Proposed change

In server 2.15 a new backup format will store backups in a tar.s2 file one file per message.

This gives us a change to edit the file along several axis:

 * Last `n` per subject
 * Delete all KV purge/delete markers which will remove dead keys from KV storage
 * Messages per subject since a certain time
 * Messages matching certain subjects and patterns
 * Anonymise data in the backup keeping subjects and header keys the same, perhaps even change subjects along with a lookup table

### Use case

Ability to compact or extract new streams from backups.

Improving the support story so we can recreate customer issues without getting access to customer data

ploubser:
# Stream backup/restore v2 client 

## Investigation Topics

### Stream edits

With messages in backups now written as single files, we want to be able to filter these messages and create a new backup,
containing only messages that pass a filter.

The filters fall into two classes, and the class is what determines the shape of the algorithm (see Implementation):

- **Stateless** — the keep/drop decision for a message is a pure function of that message alone, so it can be made the
  moment the message is read. These are the filters we tested:
  - **Subject** — match the message subject against one or more subjects/wildcards (e.g. `orders.>`). Has an inverted
    form to exclude matches instead.
  - **Header** — match a header name to a specific value, or match simply on a header being present. Both have inverted
    forms to exclude.
  - **Time** — keep messages whose timestamp falls after a `since` bound and/or before a `before` bound.
  - **Payload** — match the message body by substring or by regular expression. Both have inverted forms to exclude.
- **Per-subject (deferred)** — the decision for a message depends on the other messages for the same subject, so it
  cannot be made until the whole stream has been read:
  - **Last-N per subject** — keep only the most recent N messages of each subject.
  - **KV compaction (`kv_compact`)** — see below.


Independently of which filters are applied, the output is always a new backup. The surviving messages are
renumbered from sequence 1 with deleted-sequence gaps removed, consumers are dropped, and the stream state is reset.
Only the message count and first/last sequence are written into the new `state.json`. The server recomputes byte counts, 
subject counts and first/last timestamps from the restored messages. Each message also keeps its original timestamp.

`kv_compact` is a special case. It reproduces how a KV bucket looks after server-side compaction: for each subject
(key) it keeps only the latest message, and additionally drops the subject's history entirely when that latest
message is a delete/purge tombstone. It is special in two ways: the decision is deferred, and in the tombstone case the key's whole history is removed rather
than keeping a single message.

[...]

PR #1725, opened 2026-09-09
(#1676) Add nats backup command for editing stream backups on disk

ploubser:
Adds nats backup edit, info and validate, working on the directory written by nats stream backup. Only 2.15 backups are supported, older formats are refused.

Tested end to end against nats-server 2.10 and 2.15.0-preview.1. stream, kv and account backup and restore round trip on both servers, a 2.10 backup restores into a 2.15 server, and restoring a 2.15 backup into a 2.10 server fails without leaving a partial stream.


ploubser:
@codex review

cli/cheats/backup.md
```diff
@@ -0,0 +1,26 @@
+# Checking a backup taken with nats stream backup
+nats backup validate ./orders-backup
```
ripienaar, line 2: need to think about this one, its unfortunate `nats s backup` is where it is, ultimately be good to have a single place for all this.

Something like:

`nats backup stream`
`nats backup account`
`nats backup edit` etc

and then we deprecate via warnings `nats s backup` and `nats account backup` and remove those next release

later once we get some other backup options here for things that can be backed up then this is a natural place to focus on

This backup format also opens a few other interesting options to explore, now to backup just one subject that might be 100kb in a 100GB stream you have to backup the whole thing and then edit, lame.

Instead we can have:

```
# a filtering consumer to backup and make a new stream with just those messages
$ nats backup consumer STREAM TARGET <consumer create flags> --context prod
$ nats backup restore SOURCE --stream OTHER --context staging

# core nats sub into a custom stream
$ nats backup sub STREAM TARGET subjects <stream create flags>
```

or `nats backup stream` can gain flags that have us instead walk the thing with batch direct get pulling out last per subject over a wildcard.  Since we can now just make up backups from scratch

ripienaar (changes requested):

cli/backup_command.go
```diff
@@ -0,0 +1,349 @@
...
+	edit.Flag("exclude-payload-match", "Drop messages whose payload matches a regular expression").PlaceHolder("REGEX").StringsVar(&c.excludePayloadMatch)
+	edit.Flag("last-per-subject", "Keep only the newest N messages of every subject").PlaceHolder("N").IsSetByUser(&c.lastPerSubjectSet).IntVar(&c.lastPerSubject)
+	edit.Flag("kv-compact", "Keep the latest revision of each key and drop deleted keys. KV backups only, exclusive with --last-per-subject").UnNegatableBoolVar(&c.kvCompact)
+	edit.Flag("renumber", "Number the kept messages from 1 and drop all consumers").UnNegatableBoolVar(&c.renumber)
+	edit.Flag("dry-run", "Only shows what the edit would do, does not write the target").UnNegatableBoolVar(&c.dryRun)
+	edit.Flag("obfuscate", "Replace names and subjects with keyed hashes, drop message bodies. Keeps counter stream values. Writes the key file beside the target").UnNegatableBoolVar(&c.obfuscate)
+	edit.Flag("obfuscation-key", "Reuse the secret and mappings from an earlier key file").PlaceHolder("FILE").StringVar(&c.obfuscationKey)
+
+	info := bk.Command("info", "Stream backup information").Action(c.infoAction)
```
ripienaar, line 87: how hard would It be to add --subjects here and list subjects found in the archive like `nats s subjects` does?

cli/backup_command.go
```diff
@@ -0,0 +1,349 @@
...
+	edit.Flag("exclude-header", "Drop messages carrying a header").PlaceHolder("NAME").StringsVar(&c.excludeHeaders)
+	edit.Flag("payload-match", "Keep messages whose payload matches a regular expression").PlaceHolder("REGEX").StringsVar(&c.payloadMatch)
+	edit.Flag("exclude-payload-match", "Drop messages whose payload matches a regular expression").PlaceHolder("REGEX").StringsVar(&c.excludePayloadMatch)
+	edit.Flag("last-per-subject", "Keep only the newest N messages of every subject").PlaceHolder("N").IsSetByUser(&c.lastPerSubjectSet).IntVar(&c.lastPerSubject)
+	edit.Flag("kv-compact", "Keep the latest revision of each key and drop deleted keys. KV backups only, exclusive with --last-per-subject").UnNegatableBoolVar(&c.kvCompact)
+	edit.Flag("renumber", "Number the kept messages from 1 and drop all consumers").UnNegatableBoolVar(&c.renumber)
+	edit.Flag("dry-run", "Only shows what the edit would do, does not write the target").UnNegatableBoolVar(&c.dryRun)
+	edit.Flag("obfuscate", "Replace names and subjects with keyed hashes, drop message bodies. Keeps counter stream values. Writes the key file beside the target").UnNegatableBoolVar(&c.obfuscate)
```
ripienaar: I dont recall what we said about bodies to be honest, no bodies probably ok but add a flag to just replace it with len(body)*"X" so it compress well etc

cli/backup_command.go
```diff
@@ -0,0 +1,349 @@
...
+	validate := bk.Command("validate", "Validates that a stream backup is complete and restorable").Action(c.validateAction)
+	validate.Tag("scope:user", "impact:ro")
+	validate.Arg("source", "The directory holding the backup").Required().ExistingDirVar(&c.source)
+}
+
+func init() {
+	registerCommand("backup", 1, configureBackupCommand)
+}
+
+func parseBackupTime(s string) (time.Time, error) {
```
ripienaar: attach to the struc

cli/backup_command.go
```diff
@@ -0,0 +1,349 @@
...
+	if err != nil {
+		return err
+	}
+
+	c.showEditResult(res)
+
+	return nil
+}
+
+func (c *backupCmd) showEditResult(res *backup.Result) {
```
ripienaar, line 211: be good to see subject count before and after also

cli/backup_command.go
```diff
@@ -0,0 +1,349 @@
...
+	renumber            bool
+	dryRun              bool
+	obfuscate           bool
+	obfuscationKey      string
+}
+
+func configureBackupCommand(app commandHost) {
+	c := &backupCmd{}
+
+	bk := app.Command("backup", "Inspect and edit stream backups on disk")
```
ripienaar, line 62: need a utility here to look up from the keys file

cli/backup_command.go
ripienaar: we might also need to think about padding these keys I have a header value `dtm51j0d6qo3g7hd` that's sweet but the real value is `ORDERS_15 29515852 > > js.in.orders_15 WVoib1Oo` and when debugging why streams are doing weird things these lengths matter so we could pad

something similar with subjects happening `js.in.orders_96` becomes something like `6i55j8cb06hk8te9.s2h78g1ues7vkuct.8lf09vkn3ajcj8eu` which while fine does change the shape of the backup significantly here though we can't do anything as if we trim the hashes aggressively we'd end up with collisions between entirely different subjects - we can pad though for longer ones?

cli/backup_command.go
```diff
@@ -0,0 +1,349 @@
...
+	rep, err := backup.Verify(c.source)
+	if err != nil {
+		return err
+	}
+
+	seqs := "no messages"
+	if rep.Messages > 0 {
+		seqs = fmt.Sprintf("sequences %s to %s", f(rep.FirstSeq), f(rep.LastSeq))
+	}
+	fmt.Printf("OK: %s entries, %s consumers, %s messages, %s\n", f(rep.Entries), f(rep.Consumers), f(rep.Messages), seqs)
```
ripienaar: show subject count if possible

cli/cheats/backup.md
ploubser: Yeah, this is a good idea. As you mentioned offline (putting it here for completeness) is that we should rename the backup output file too. Not part of this PR but I'll circle back.

cli/backup_command.go
ploubser: Added. Looks the same as `nats s subjects`

cli/backup_command.go
ploubser: I agree it's better to 0 the body instead of dropping it. The size of the data matters if this is a diagnostics tool. Updated to do that instead and we're now also padding the larger obfuscated strings to match the size.

cli/backup_command.go
ploubser: Done

cli/backup_command.go
ploubser: Added

cli/backup_command.go
ploubser: Added `backup lookup` to let users lookup the origin value.

ploubser:
@codex review

cli/backup_command.go
```diff
@@ -0,0 +1,394 @@
...
+	edit.Flag("exclude-payload-match", "Drop messages whose payload matches a regular expression").PlaceHolder("REGEX").StringsVar(&c.excludePayloadMatch)
+	edit.Flag("last-per-subject", "Keep only the newest N messages of every subject").PlaceHolder("N").IsSetByUser(&c.lastPerSubjectSet).IntVar(&c.lastPerSubject)
+	edit.Flag("kv-compact", "Keep the latest revision of each key and drop deleted keys. KV backups only, exclusive with --last-per-subject").UnNegatableBoolVar(&c.kvCompact)
+	edit.Flag("renumber", "Number the kept messages from 1 and drop all consumers").UnNegatableBoolVar(&c.renumber)
+	edit.Flag("dry-run", "Only shows what the edit would do, does not write the target").UnNegatableBoolVar(&c.dryRun)
+	edit.Flag("obfuscate", "Replace names and subjects with keyed hashes and message bodies with zeros of the same length. Keeps counter stream values. Writes the key file beside the target").UnNegatableBoolVar(&c.obfuscate)
+	edit.Flag("obfuscation-key", "Reuse the secret and mappings from an earlier key file").PlaceHolder("FILE").StringVar(&c.obfuscationKey)
+
+	info := bk.Command("info", "Stream backup information").Action(c.infoAction)
```
ripienaar, line 87: we get a quite long pause here on bigger backups with no feedback, ideally we add in some progress bars here but thats quite a thing, so maybe just print some status that its checking?

But for sure the API I going to need some kind of progress bar for some of these long operations

Also on validate and edit if possible (though that could be hard because of the multi passes I suppose?)

cli/backup_command.go
```diff
@@ -0,0 +1,394 @@
...
+	cols.AddRowUnlimitedIf("Maximum Message Size", humanize.IBytes(uint64(max(int64(cfg.MaxMsgSize), 0))), cfg.MaxMsgSize == -1)
+	cols.AddRowUnlimited("Maximum Consumers", int64(cfg.MaxConsumers), -1)
+
+	cols.AddSectionTitle("Consumers")
+	cols.AddRow("Count", len(nfo.Consumers))
+	cols.AddRowIf("Names", nfo.Consumers, len(nfo.Consumers) > 0)
+
+	cols.AddSectionTitle("Messages")
+	cols.AddRow("Messages", nfo.Messages)
+	cols.AddRow("Subjects", nfo.NumSubjects)
```
ripienaar, line 318: move this subjects to the last one in this section so visually this and the next section looks similar up to the last item, that'll help when scanning the output

cli/backup_command.go
```diff
@@ -0,0 +1,394 @@
...
+	obfuscationKey      string
+	listSubjects        bool
+	keyFile             string
+	values              []string
+}
+
+func configureBackupCommand(app commandHost) {
+	c := &backupCmd{}
+
+	bk := app.Command("backup", "Inspect and edit stream backups on disk")
```
ripienaar, line 62: have to make `stream backup` detect new format and name the files not tar.s2

ripienaar (changes requested):

cli/backup_command.go
ploubser: Good shout, added it to everything.

cli/backup_command.go
ploubser: 👍

cli/backup_command.go
ploubser: Agree, will do it as a separate PR.

ploubser:
@codex review the new changes.

ripienaar (approved):

PR #1734, opened 2026-09-15
feat: add `auth nuid` command

joeriddles:
Add command to generate [NUIDs](https://github.com/nats-io/nuid), similar to `auth nkey`.

```shell
❯ nats auth nuid gen
4MalF4es5klgiRBugB15Gg

❯ nats auth nuid --help
usage: nats auth nuid <command> [<args> ...]

Create NUIDs

Subcommands:
  auth nuid gen  Generates NUIDs
```


cli/auth_nuid_command.go
```diff
@@ -0,0 +1,37 @@
...
+)
+
+type authNuidCommand struct{}
+
+func configureAuthNuidCommand(auth commandHost) {
+	c := &authNuidCommand{}
+
+	nu := auth.Command("nuid", "Create NUIDs")
+
+	nuGen := nu.Command("gen", "Generates NUIDs").Action(c.genAction)
```
ripienaar, line 30: not adding the validate command?

joeriddles: I omitted for now since nuid repo only has `func IsValid(s string) bool`. Seems like a poor CLI experience to know that it's invalid but not why.

My plan is to add something like `func Parse(s string) (..., error)` to the nuid repo and add the validate command here after that merges.

ripienaar (approved):

ripienaar:
Thank you @joeriddles

Issue #1718, opened 2026-09-03
nats sub - feedback when msgs are dropped


borjaetxebarria:
### Proposed change

In a simple test that floods a `nats-cli` subscriber, for example:

```bash
nats sub foo
nats bench pub foo --msgs=1000000
```

`nats sub` prints the received messages along with a `[#counter]`. Under heavy load, that counter will show fewer messages than were actually published (e.g. less than the full 1M).

This is expected: `nats-cli` intentionally drops messages that it cannot process/render to the terminal fast enough. That behavior is reasonable.

When `--trace` is enabled, these drops are reported as they occur. 

So in normal operation, the errors are hidden, possibly intentionally to avoid cluttering the terminal.

The problem is that, without any indication that messages were dropped by the CLI, users may interpret the lower message count as message loss in `nats-cli` or even in NATS itself, raising unnecessary reliability concerns.

Some possible improvements:

* Report message-drop errors even when `--trace` is not enabled.
* Avoid per-message errors, but print a final warning indicating how many messages were dropped.
* Keep the running `[#counter]` based on messages received rather than messages rendered, so it reflects the true number of messages delivered to the subscriber even when some are not printed.

**Note:** According to Claude (*ehem...*), this may actually be a bug rather than entirely intentional behavior: duplicate `nats.ErrorHandler` options appear to be registered in `cli/util.go` (around lines 275 and 283).

### Use case

improvement

ripienaar:
Duplicate err handler is certainly a thing to fix, I think I like the keep count and report approach rather than make it worse by printing logs for every message we already couldn't print.

ripienaar:
https://github.com/nats-io/natscli/pull/1738

PR #1738, opened 2026-09-16
(#1718) Fix error handling and msg count in nats sub command

ploubser:
natsOpts registered nats.ErrorHandler twice and nats.go keeps only the last one, so the unconditional handler was dead and every async error was silent without --trace. Merge them into one handler that always logs, except slow consumer errors which stay behind --trace.

The [#N] counter in nats sub now also includes messages the client dropped.


ploubser:
@codex review

ploubser:
@codex review

ripienaar (approved):

PR #1740, opened 2026-09-18
Meta rescue with single domain

MauriceVanVeen:
Clarify that `nats server cluster rescue` requires an explicitly set domain, even if only a single one is used. Otherwise the CLI would only report `no compatible servers found`.


cli/server_cluster_command.go
```diff
@@ -174,15 +174,17 @@ func (c *SrvClusterCmd) metaRescueAction(_ *fisk.ParseContext) error {
...
 			names := iu.MapKeys(domains)
 			slices.Sort(names)
 			for i, d := range names {
 				if d == "" {
 					names[i] = "no domain"
 				}
 			}
-
+			if len(names) == 1 {
+				return fmt.Errorf("domain found, explicitly set it to %s", names[0])
```
ripienaar, line 186: hmm, this seem super unhelpful if I saw this message it would annoy me, bit like "foo --help not supported please run foo help".

Let me think about it, maybe we just set it and continue as if it was set.

MauriceVanVeen: Could change it to optimistically match everything if `domain` is empty, but if we then notice that there are more than one domain you need to explicitly pick one.

A system with no domain or one with only one would continue without needing to specify it. However, we should probably still use the explicit domain in the rescue request in that case.

ripienaar: Exactly what I was getting at yes, if we see a domain we communicate then we take that one and ensure we include it in th request to avoid the situation with missing some slow responses.

Let me give it some thought but thats where I was leaning also

